PR #3, opened 2020-10-22
Enable tcmalloc to improve the malloc performance

guzongmin:
Replace the dlmalloc with tcmalloc


guzongmin:
The tcmalloc could only works under HW. SIM mode broken is expected.

guzongmin:
replaced by https://github.com/occlum/ngo/pull/180

PR #27, opened 2021-07-09
[async-rt] Remove sched_callback.

ShuochengWang:
#26 [async-rt] Should remove set_sched_callback? 

- remove sched_callback in async-rt config and executor.
- add background task (io_uring poll function) when init async-rt.
    - need shutdown these tasks before shutdown executor (not implemented yet).
    - consider add attribute / status to `task` struct later.
- use yield() in background task to avoid running too long. 
    - (one insight: without yield, io_uring poll function will always running.)

needs to do:

- distinguish background tasks and user tasks.    (add attributes for `task` struct ?)
- select suitable thread for background tasks.
- add status for task 
      - avoid waking up repeatedly. 
      - do not wake up tasks with shutdown-status. 
- shutdown background tasks in executor.shutdown().

### DIFF
--- a/src/libos/crates/async-rt/src/config.rs
+++ b/src/libos/crates/async-rt/src/config.rs
@@ -8,15 +8,6 @@ pub fn set_parallelism(parallelism: u32) {
     CONFIG.set_parallelism(parallelism);
 }
 
-/// Set a callback function that will be invoked right before a task will be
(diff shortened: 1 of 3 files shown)


tatetian (approved):
Looks good. Thanks for the contribution!

PR #28, opened 2021-07-09
Add cancel and timeout for io-uring-callback.

ShuochengWang:
#### add do_timeout for libos.

- To implement cancel, we pass a IoUring reference parameter for cancel().   (It's complicated to store IoUring ref in the IoToken, which causes circular-reference, lifetime problem ...)
- Add do_timeout for libos using io-uring timeout.
- Only use do_timeout() in do_sigtimedwait() for now. Since io-uring timeout cannot support signal wake-up and remaining time (nanosleep)

##### How to support nanosleep?
- require:  wake up by signal
- require:  calculate remaining time if waked up by signal.

### DIFF
--- a/src/libos/crates/io-uring-callback/src/io_handle.rs
+++ b/src/libos/crates/io-uring-callback/src/io_handle.rs
@@ -13,20 +13,23 @@ cfg_if::cfg_if! {
 
 /// The handle to an I/O request pushed to the submission queue of an io_uring instance.
 #[derive(Debug)]
-pub struct IoHandle(Arc<IoToken>);
+pub struct IoHandle(pub(crate) Arc<IoToken>);
 
 /// The state of an I/O request represented by an [`IoHandle`].
+/// If a request is in `Processed` or `Cancelled` state, means that the request is completed.
 #[derive(Debug, Copy, Clone, PartialEq, Eq)]
 pub enum IoState {
-    /// Submitted.
+    /// The I/O request has been submitted.
     Submitted,
-    /// Completed with a return value.
-    Completed(i32),
-    /// Cancelling (not used yet).
+    /// The I/O request has been processed by the kernel and returns a value.
+    Processed(i32),
+    /// The I/O request is being cancelled.
     Cancelling,
-    /// Cancelled (not used yet).
+    /// The I/O request has been cancelled by the kernel.
     Cancelled,
 }
+
+const CANCEL_RETVAL: i32 = -libc::ECANCELED;
 
 impl IoHandle {
     pub(crate) fn new(token: Arc<IoToken>) -> Self {
@@ -42,13 +45,6 @@ impl IoHandle {
     pub fn retval(&self) -> Option<i32> {
         self.0.retval()
     }
-
-    /// Cancel the I/O request.
-    ///
-    /// This is NOT implemented, yet.
-    pub fn cancel(&self) {
-        self.0.cancel()
-    }
 }
 
 impl Unpin for IoHandle {}
@@ -59,13 +55,11 @@ impl Future for IoHandle {
     fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
         let mut inner = self.0.inner.lock().unwrap();
         match inner.state {
-            IoState::Completed(retval) => Poll::Ready(retval),
-            IoState::Submitted => {
+            IoState::Processed(retval) => Poll::Ready(retval),
+            IoState::Cancelled => Poll::Ready(CANCEL_RETVAL),
+            IoState::Submitted | IoState::Cancelling => {
                 inner.waker = Some(cx.waker().clone());
                 Poll::Pending
-            }
-            _ => {
-                todo!("handle cancel-related states");
             }
         }
     }
@@ -73,8 +67,8 @@ impl Future for IoHandle {
 
 impl Drop for IoHandle {
     fn drop(&mut self) {
-        // The user cannot drop a handle without completing or canceling it first.
-        assert!(matches!(self.state(), IoState::Completed(_) | IoState::Cancelled));
+        // The user cannot drop a handle if the request isn't completed.
+        assert!(matches!(self.state(), IoState::Processed(_) | IoState::Cancelled));
     }
 }
 
@@ -87,8 +81,8 @@ pub(crate) struct IoToken {
 }
 
 impl IoToken {
-    pub fn new(callback: impl FnOnce(i32) + Send + 'static) -> Self {
-        let inner = Mutex::new(Inner::new(callback));
+    pub fn new(completion_callback: impl FnOnce(i32) + Send + 'static, token_key: u64) -> Self {
+        let inner = Mutex::new(Inner::new(completion_callback, token_key));
         Self { inner }
     }
 
@@ -113,9 +107,11 @@ impl IoToken {
         (callback)(retval);
     }
 
-    pub fn cancel(&self) {
+    /// Change the state from submited to cancelling.
+    /// If transition succeeds, return the token_key for following cancel operation.
+    pub fn transit_to_cancelling(&self) -> Result<u64, ()> {
         let mut inner = self.inner.lock().unwrap();
-        inner.cancel()
+        inner.transit_to_cancelling()
     }
 }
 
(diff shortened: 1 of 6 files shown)


ShuochengWang:
#24 Add timeout support
#28 [io-uring-callback] Add cancel to io_uring_callback

src/libos/crates/io-uring-callback/src/lib.rs
```diff
@@ -388,6 +392,30 @@ impl IoUring {
         }
     }
 
+    /// Wait new I/O completions in the completions queue of io_uring.
+    ///
+    /// Upon receiving completed I/O, the corresponding user-registered callback functions
+    /// will get invoked and the `IoHandle` (as a `Future`) will become ready.
+    pub fn wait_completions(&self) {
```
tatetian, line 433: If `poll_completions` returns the number of completions, then `wait_completions` would be simply equivalent to
```rust
while io_uring.poll_completions() == 0 {}
```

src/libos/crates/io-uring-callback/src/io_handle.rs
```diff
@@ -44,10 +48,13 @@ impl IoHandle {
...
-    pub fn cancel(&self) {
-        self.0.cancel()
+    pub fn cancel(&self, io_uring: &IoUring) {
+        self.0.cancel(io_uring)
+    }
+
+    /// returns the user_data of the I/O request.
+    pub fn user_data(&self) -> u64 {
+        self.0.user_data()
     }
```
tatetian, line 58: Do not expose `user_data` through `IoHandle`. `user_data` has been used by `IoUring` internally. The user of `IoUring` or `IoHandle` should never have to deal with `user_data`.

src/libos/crates/io-uring-callback/src/io_handle.rs
```diff
@@ -44,10 +48,13 @@ impl IoHandle {
...
 
     /// Cancel the I/O request.
-    ///
-    /// This is NOT implemented, yet.
-    pub fn cancel(&self) {
-        self.0.cancel()
+    pub fn cancel(&self, io_uring: &IoUring) {
```
tatetian, line 51: This interface is not user friendly and safe (what if the `io_uring` is not the one that generates the `IoHandle`? ). I cannot think of a better way for now, but I believe we can do better.

src/libos/crates/io-uring-callback/src/io_handle.rs
```diff
@@ -87,8 +94,8 @@ pub(crate) struct IoToken {
 }
```
tatetian, line 94: Canceling is different from normal I/O in that
1. It does not make sense to hold a handle for cancel;
2. It does not make sense to maintain state for cancel;
3. It does not make sense to add callback (the callback you write is essentially empty) for cancel.

For the reasons above, I think we should extend the notion of `IoToken`, with which we can differentiate between normal I/O operations and cancel operations.
```rust
pub enum IoToken {
    Request(IoRequest),
    Cancel,
}

pub struct IoRequest {
    state: IoState,
    callback: Option<Callback>,
    waker: Option<Waker>,
}

pub struct IoHandle(Arc<IoRequest>);
```

src/libos/crates/io-uring-callback/src/lib.rs
```diff
@@ -351,6 +351,40 @@ impl IoUring {
...
+    /// See the safety section of the `IoUring`.
+    pub unsafe fn timeout_remove(
+        &self,
+        user_data: u64,
+        flags: types::TimeoutFlags,
+        callback: impl FnOnce(i32) + Send + 'static,
+    ) -> IoHandle {
+        let entry = opcode::TimeoutRemove::new(user_data).flags(flags).build();
+        self.push_entry(entry, callback)
+    }
```
tatetian, line 386: What is the relationship between cancel and timeout_remove? Can we unify cancel and remove with `IoHandle::cancel`?

src/libos/src/signal/do_sigtimedwait.rs
```diff
@@ -12,9 +12,9 @@ pub async fn do_sigtimedwait(interest: SigSet, timeout: Option<&Duration>) -> Re
         "do_rt_sigtimedwait: interest: {:?}, timeout: {:?}",
         interest, timeout,
     );
-    // TODO: support timeout
+
     if let Some(timeout) = timeout {
-        warn!("do not support timeout yet");
+        crate::time::do_timeout(timeout);
```
tatetian, line 17: I don't think this commit gives set a good example of how timeout can be supported in general. There must be some kind of more high-level abstraction that supports waiting with a timeout, instead of using io_uring's timeout directly.

src/libos/crates/io-uring-callback/src/lib.rs
ShuochengWang:
``` Rust
/// Return the number of I/O completions.
pub fn poll_completions(&self) -> usize {}

/// Wait for at least min_complete new I/O completions in the completions queue of io_uring.
/// Return the number of I/O completions.
pub fn wait_completions(&self, min_complete: usize) -> usize {
    // implemented by self.poll_completions()
}
```

src/libos/crates/io-uring-callback/src/io_handle.rs
ShuochengWang: However, io_uring OP, such as, poll_remove, timeout_remove, async_cancel, etc..., need user_data arg.

If we do not expose `user_data` through IoHandle, io-uring-callback can not privide some io-uring operations directly to users. But we can privide these operations in higher level abstraction, like,  `handle.cancel()`

src/libos/crates/io-uring-callback/src/io_handle.rs
ShuochengWang: I agree.

I use this interface now because I haven't thought a good way to get io-uring instance in the IoToken.

src/libos/crates/io-uring-callback/src/io_handle.rs
ShuochengWang: I agree.

Canceling in io-uring is different from other operations: **We don't have to wait for it to complete**.

- If canceling successes later, it's ok.
- If canceling fails later, it's still ok, since the operation itself may not support cancellation or has been completed.

src/libos/crates/io-uring-callback/src/lib.rs
ShuochengWang: - **async_cancel:** general cancel (from kernel 5.5)
- **timout_remove:** cancel timeout (from kernel 5.5), update timeout (from kernel 5.11)

In cancel cases, `timeout_remove` is a subset of `async_cancel`, hence we can use `async_cancel` in `IoHanlde::cancel`.

src/libos/src/signal/do_sigtimedwait.rs
ShuochengWang: Interface examples:

``` Rust
let timer = Timer::new(io_uring);
let firer = timer.firer();
...
timer.sleep(duration).await
...
timer.sleep_forever().await
...
// firer.fire() ? 
// Should we guarantee that the cancel will be finished in this case?
firer.fire().await
```

``` Rust
if let Err(_) = timeout(duration, future).await {
    println!("the future did not complete within the duration");
}
```

ShuochengWang: 1. Should we support Instant interface in timer? e.g., `timer.sleep_at(instant)`.
2. nanosleep(), how to wake by signal?

src/libos/crates/io-uring-callback/src/io_handle.rs
ShuochengWang: Another point, we can only record IoRequests, implement canceling only by a inner interface (no token).

maybe we need two cancel interface?

- non-block cancel: just send cancel request, no wait
- block cancel: send and wait (await)

src/libos/crates/io-uring-callback/src/lib.rs
ShuochengWang: `busywait_completions`

src/libos/crates/io-uring-callback/src/io_handle.rs
```diff
@@ -22,10 +22,10 @@ pub enum IoState {
     Submitted,
     /// Completed with a return value.
     Completed(i32),
-    /// Cancelling (not used yet).
+    /// Cancelling.
     Cancelling,
-    /// Cancelled (not used yet).
-    Cancelled,
+    /// Cancelled with a return value
+    Cancelled(i32),
```
tatetian, line 28: The `Cancelled` variant does not need to be associated with a return value.

I have given more explanations this in my other comments about the usage of the cancel method. But one simple arguement is that even the `Cancelled` variant has an attached return value, it would have only one possible value---`-libc::ECANCELED`---according to the logic in the `complete` method. So keeping the return value for `Cancelled` is meaningless.

src/libos/crates/io-uring-callback/src/lib.rs
```diff
@@ -442,9 +481,52 @@ impl IoUring {
         io_handle
     }
 
+    /// Cancel an ongoing I/O request.
+    ///
+    /// # Safety
+    ///
+    /// See the safety section of the `IoUring`.
+    pub fn cancel(&self, handle: &IoHandle) {
+        self.inner_cancel(handle.0.clone(), true);
```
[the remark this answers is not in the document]
ShuochengWang: We can avoid it by ref.
```Rust
pub fn cancel(&self, handle: &IoHandle) {
        self.do_cancel(&handle.0);
}
fn do_cancel(&self, token: &Arc<IoToken>) {}
```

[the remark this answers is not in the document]
ShuochengWang: Well, since we don't need `cancel_all`, we can do all cancel-things in `cancel()` instead of `do_cancel`

ShuochengWang:
All above problems has been fixed in latest commit. @tatetian

src/libos/crates/io-uring-callback/src/lib.rs
```diff
@@ -375,17 +391,42 @@ impl IoUring {
     ///
     /// Upon receiving completed I/O, the corresponding user-registered callback functions
     /// will get invoked and the `IoHandle` (as a `Future`) will become ready.
-    pub fn poll_completions(&self) {
+    /// Return the number of I/O completions. The Completions of Cancel do not count in the number.
```
tatetian, line 394: Well, I think it is time to clear things up. We have three concepts that may confuse ourselves as well as users.

*  I/O io_uring completions and completion queues. The completion in this context means the request is handled by the kernel, and the request could succeed, fail, or get cancelled.
* The completion state of `IoHandle`, which is `IoState::Completed`. 
*  The completions of cancel operations, which is `IoState::Cancelled`.

The name of `poll_completions` is originally meant to use the first meaning, but the comments of "completions of cancel do not count" effectively shifts the meaning to the third one. To make the matter worse, the internal implementation and the users of `IoHandle::state` method would take the second meaning. Admittedly, this situation is not good: confused concepts lead to implementation bugs.

To address this confusion, I make the following proposal.

* Let's only talk about _completions_ in the original meaning used by io_uring literatures. That is, completion means a request is handled by the kernel, and the request could succeed, fail, or get cancelled.
* Rename `IoState::Completed` to `IoState::Processed` (any better name suggestion?).
* When referring the callback function argument, always stick with the name `completion_callback`, to emphasize the callback is called when the io_uring request is completed, including the case of being cancelled.
* Fix the implementation of `poll_completions` to count cancellations and fix the docs.
* Fix other docs or add more docs accordingly.

tatetian (commented):
Good job on improvements. I think we are close to finish this PR. There are still some minor issues.

By the way, could you mark conversations as resolved if they are resolved in the future? This would make my life as a reviewer easier. Thanks.

Issue #29, opened 2021-07-20
[RFC] Timeout support


tatetian:
# Motivation

It is common to have event waiting logics bounded with timeouts. But currently, all NGO code will wait forever if no interesting events arrive. And since there are so many system calls that may wait with timeout, we need to figure out a general mechanism to support timeout.

# Background

Currently, all event wait/wakeup code in NGO eventually relies on `Waiter`/`WaiterQueue` provided by the `async-rt` crate.  And a convenient macro named `waiter_loop` is provided to make it even easier for the most common pattern when using `Waiter`/`WaiterQueue`, which is to 1) try to make progress with a task, 2) if not ready, wait for notifications before trying again.

Here are some examples.

## Exit/Wait4

The exit and wait4 system calls use `Waiter` and `WaiterQueue`.

```rust
// File: libos/process/do_wait4.rs

pub async fn do_wait4(child_filter: &ProcessFilter) -> Result<(pid_t, i32)> {
    let thread = current!();
    let process = thread.process();

    waiter_loop!(process.exit_waiters(), {
        // Lock order: always lock parent then child to avoid deadlock
        let mut process_inner = process.inner();

        let unwaited_children = /* get all possible children */

        if unwaited_children.len() == 0 {
            return_errno!(ECHILD, "Cannot find any unwaited children");
        }

        // Return immediately if a child that we wait for has already exited
        let zombie_child = unwaited_children
            .iter()
            .find(|child| child.status() == ProcessStatus::Zombie);
        if let Some(zombie_child) = zombie_child {
            let zombie_pid = zombie_child.pid();
            let exit_status = free_zombie_child(&process, process_inner, zombie_pid);
            return Ok((zombie_pid, exit_status));
        }	
	}
}

// File: libos/process/do_exit.rs

fn exit_process(thread: &ThreadRef, term_status: TermStatus) {
	// ...
	
    // Notify the parent that this child process's status has changed
    parent.exit_waiters().wake_all();
	
	// ...
}
```

## Sigtimedwait

The sigtimedwait also uses `Waiter` and `WaiterQueue` to wait for signals.

```rust
pub async fn do_sigtimedwait(interest: SigSet, timeout: Option<&Duration>) -> Result<siginfo_t> {
    if let Some(timeout) = timeout {
        warn!("do not support timeout yet");
    }

    let thread = current!();
    let process = thread.process().clone();

    // Interesting, blocked signals
    let interest = {
        let blocked = thread.sig_mask().read().unwrap();
[...]

PR #31, opened 2021-07-29
Timeout support.

ShuochengWang:
#29 [RFC] Timeout support

- [async-rt]
    - Add timeout feature to the async-rt.
- [ngo]
    - Add rust-hash-wheel-timer submodule. Port rust-hash-wheel-timer crate to rust-sgx-sdk. (https://github.com/ShuochengWang/rust-hash-wheel-timer). 
- [libos]
    - Support timeout in do_futex / do_sigtimedwait
    - Support nanosleep

### DIFF
--- a/src/libos/crates/async-rt/src/executor.rs
+++ b/src/libos/crates/async-rt/src/executor.rs
@@ -2,7 +2,6 @@ use flume::{Receiver, Sender};
 use futures::task::waker_ref;
 
 use crate::config::CONFIG;
-#[cfg(feature = "thread_sleep")]
 use crate::parks::Parks;
 use crate::prelude::*;
 use crate::sched::Affinity;
@@ -33,14 +32,13 @@ pub(crate) struct Executor {
     task_senders: Vec<Sender<Arc<Task>>>,
     next_run_queue_id: AtomicU32,
     is_shutdown: AtomicBool,
-    #[cfg(feature = "thread_sleep")]
     parks: Parks,
 }
 
 impl Executor {
     pub fn new(parallelism: u32) -> Result<Self> {
         if parallelism == 0 {
-            return Err("invalid argument");
+            return_errno!(EINVAL, "invalid argument");
         }
 
         const MAX_QUEUED_TASKS: usize = 1_000;
@@ -54,8 +52,6 @@ impl Executor {
 
         let is_shutdown = AtomicBool::new(false);
         let next_run_queue_id = AtomicU32::new(0);
-
-        #[cfg(feature = "thread_sleep")]
         let parks = Parks::new(parallelism);
 
         let new_self = Self {
@@ -64,7 +60,6 @@ impl Executor {
             task_senders,
             next_run_queue_id,
             is_shutdown,
-            #[cfg(feature = "thread_sleep")]
             parks,
         };
         Ok(new_self)
@@ -88,13 +83,10 @@ impl Executor {
 
                 match task_res {
                     Err(_) => {
-                        #[cfg(feature = "thread_sleep")]
                         self.parks.park_timeout(
                             run_queue_id as usize,
                             core::time::Duration::from_millis(10),
                         );
-                        #[cfg(not(feature = "thread_sleep"))]
-                        core::sync::atomic::spin_loop_hint();
                         continue;
                     }
                     Ok(task) => task,
@@ -140,7 +132,6 @@ impl Executor {
             .send(task)
             .expect("too many tasks enqueued");
 
-        #[cfg(feature = "thread_sleep")]
         self.parks.unpark(thread_id);
     }
 
@@ -160,7 +151,6 @@ impl Executor {
     pub fn shutdown(&self) {
         self.is_shutdown.store(true, Ordering::Relaxed);
 
-        #[cfg(feature = "thread_sleep")]
         self.parks.unpark_all();
     }
 
--- a/src/libos/crates/async-rt/src/time/entry.rs
+++ b/src/libos/crates/async-rt/src/time/entry.rs
@@ -1,0 +1,189 @@
+use super::wheel::TIMER_WHEEL;
+use super::DURATION_ZERO;
+use crate::prelude::*;
+use core::task::Waker;
+
+/// The state of timer.
+#[derive(Debug)]
+pub enum TimerState {
+    /// This timer has not been inserted to the timer wheel.
+    Init,
+    /// This timer has been inserted to the timer wheel and waits timeout.
+    Started(StartedInner),
+    /// This timer has expired, which means the wheel has triggered the timer.
+    Expired,
+    /// This timer has been cancelled, and records elapsed time between started and cancelled state.
+    Cancelled(Duration),
+}
+
+#[derive(Debug)]
+pub struct StartedInner {
+    start_ticks: u64,
+    waker: Option<Waker>,
+}
+
+impl StartedInner {
+    pub fn new(start_ticks: u64, waker: Waker) -> Self {
+        Self {
+            start_ticks,
+            waker: Some(waker),
+        }
+    }
+
+    pub fn wake(&mut self) {
+        self.waker.take().unwrap().wake();
+    }
+
+    pub fn set_waker(&mut self, waker: Waker) {
+        self.waker = Some(waker);
+    }
+
+    pub fn elapsed_time(&self) -> Duration {
+        let elapsed_ticks = TIMER_WHEEL.latest_ticks() - self.start_ticks;
+        Duration::from_millis(elapsed_ticks)
+    }
+}
+
+/// The shared data structure of timer.
+#[derive(Debug)]
+pub struct TimerShared {
+    state: TimerState,
+    timeout: Duration,
+}
+
+/// The wrapper of `TimerShared` with `Send` and `Sync`.
+pub type TimerSharedRef = Arc<Mutex<TimerShared>>;
+
+impl TimerShared {
+    pub fn new(timeout: Duration) -> Self {
+        Self {
+            state: TimerState::Init,
+            timeout: timeout,
+        }
+    }
+
+    /// Transfer to expired state and wake up the timer.
+    pub fn fire(&mut self) {
+        match &mut self.state {
+            // This timer has been cancelled, do nothing here.
+            TimerState::Cancelled(_) => {}
+            // Transfer to expired state and wake up the timer.
+            TimerState::Started(inner) => {
+                // Since TimerShared is guarded by the lock, we can wake before setting expired state
+                inner.wake();
+                self.state = TimerState::Expired;
+            }
+            _ => panic!("Can not fire timer in init or expired state"),
+        }
+    }
+
+    /// Transfer to cancelled state
+    pub fn cancel(&mut self) {
+        match &self.state {
+            TimerState::Init => {
+                self.state = TimerState::Cancelled(DURATION_ZERO);
+            }
+            TimerState::Started(inner) => {
+                self.state = TimerState::Cancelled(inner.elapsed_time());
+            }
+            TimerState::Expired => {}
+            TimerState::Cancelled(_) => panic!("Can not cancel twice"),
+        }
+    }
+
+    /// Get the remained duration.
+    pub fn remained_duration(&self) -> Duration {
+        match &self.state {
+            TimerState::Init => self.timeout,
+            TimerState::Started(inner) => {
+                // If timeout is less than 1ms, we will wait 1ms instead.
+                // It might cause the timeout is less than elapsed_time.
+                return self.timeout.saturating_sub(inner.elapsed_time());
+            }
+            TimerState::Expired => DURATION_ZERO,
+            TimerState::Cancelled(elapsed) => self.timeout.saturating_sub(*elapsed),
+        }
+    }
+}
+
+/// The entry used by user to get remained duration.
+#[derive(Debug)]
+pub struct TimerEntry(TimerSharedRef);
+
+/// The entry used for `await` or `poll`.
+#[derive(Debug)]
+pub struct TimerFutureEntry(TimerSharedRef);
+
+/// The entry used for timer wheel. This entry will be inserted to the timer wheel.
+/// When the timer wheel trigger this entry, the wheel will invoke `fire()` to fire the timer.
+#[derive(Debug)]
+pub struct TimerWheelEntry(TimerSharedRef);
+
+impl TimerEntry {
+    pub fn new(timeout: Duration) -> Self {
+        Self(Arc::new(Mutex::new(TimerShared::new(timeout))))
+    }
+
+    // Get the remained duration.
+    pub fn remained_duration(&self) -> Duration {
+        let shared = self.0.lock();
+        shared.remained_duration()
+    }
+}
+
+impl TimerFutureEntry {
+    pub fn new(entry: &TimerEntry) -> Self {
+        Self(entry.0.clone())
+    }
+}
+
+impl Future for TimerFutureEntry {
+    type Output = ();
+
+    /// Return Ready if the timer is expired, or Pending.
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let mut shared = self.0.lock();
+        match &mut shared.state {
+            TimerState::Init => {
+                // If the timeout is zero, return directly.
+                if shared.timeout.is_zero() {
+                    shared.state = TimerState::Expired;
+                    return Poll::Ready(());
+                }
+
+                // Insert the timer entry to the timer wheel.
+                let entry = TimerWheelEntry(self.0.clone());
+                let start_ticks = TIMER_WHEEL.insert_entry(entry, shared.timeout);
+                // Transfer to started state, set start_tick and waker.
+                let inner = StartedInner::new(start_ticks, cx.waker().clone());
+                shared.state = TimerState::Started(inner);
+                Poll::Pending
+            }
+            TimerState::Started(inner) => {
+                inner.set_waker(cx.waker().clone());
+                Poll::Pending
+            }
+            TimerState::Expired => Poll::Ready(()),
+            TimerState::Cancelled(_) => panic!("Can not poll cancelled timer entry"),
+        }
+    }
+}
+
+impl Drop for TimerFutureEntry {
+    /// Make sure that the state is Init, Expired or Cancelled after drop.
+    /// If the state is Started, cancel the timer before dropped.
+    fn drop(&mut self) {
+        let mut shared = self.0.lock();
+        if let TimerState::Started(_) = shared.state {
+            shared.cancel();
+        }
+    }
+}
+
+impl TimerWheelEntry {
+    /// Fire the timer.
+    pub fn fire(&self) {
+        let mut shared = self.0.lock();
+        shared.fire();
+    }
+}
--- a/src/libos/crates/async-rt/src/time/instant.rs
+++ b/src/libos/crates/async-rt/src/time/instant.rs
@@ -1,0 +1,125 @@
+use crate::prelude::*;
+use core::ops::{Add, AddAssign, Sub, SubAssign};
+use vdso_time::{clock_gettime, ClockId};
+
+lazy_static! {
+    static ref LAST_NOW: Mutex<Instant> = Mutex::new(Instant(DURATION_ZERO));
+}
+
+pub const DURATION_ZERO: Duration = Duration::from_nanos(0);
+
+/// A measurement of a monotonically nondecreasing clock. Opaque and useful only with Duration.
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Ord, PartialOrd)]
+pub struct Instant(pub(crate) Duration);
+
+impl Instant {
+    /// Returns an instant corresponding to “now”.
+    ///
+    /// Rust std says that the clock of linux + x86 environment is real-monotonic.
+    /// When sgx feature enabled, we are in linux + x86 env, the clock should be real-monotonic.
+    /// However, the clock in sgx env is untrusted, we can not guarantee that clock is real-monotonic.
+    /// To ensure the instant is monotonically nondecreasing, We keep a global "latest now" instance
+    /// which is returned instead of what the OS says if the OS goes backwards.
+    pub fn now() -> Self {
+        let os_now = Instant(clock_gettime(ClockId::CLOCK_MONOTONIC).unwrap());
+
+        let mut last_now = LAST_NOW.lock();
+        let now = core::cmp::max(*last_now, os_now);
+        *last_now = now;
+
+        now
+    }
+
+    /// Returns the amount of time elapsed from another instant to this one.
+    pub fn duration_since(&self, earlier: Instant) -> Duration {
+        self.0
+            .checked_sub(earlier.0)
+            .expect("earlier is later than self.")
+    }
+
+    /// Returns the amount of time elapsed from another instant to this one,
+    /// or None if that instant is later than this one.
+    pub fn checked_duration_since(&self, earlier: Instant) -> Option<Duration> {
+        self.0.checked_sub(earlier.0)
+    }
+
+    /// Returns the amount of time elapsed from another instant to this one,
+    /// or zero duration if that instant is later than this one.
+    pub fn saturating_duration_since(&self, earlier: Instant) -> Duration {
+        self.0.checked_sub(earlier.0).unwrap_or(DURATION_ZERO)
+    }
+
+    /// Returns the amount of time elapsed since this instant was created,
+    /// or zero duration if Instant::now() is earlier than this one.
+    pub fn elapsed(&self) -> Duration {
+        Instant::now() - *self
+    }
+
+    /// Returns Some(t) where t is the time self + duration if t can be represented as Instant
+    /// (which means it’s inside the bounds of the underlying data structure), None otherwise.
+    pub fn checked_add(&self, duration: Duration) -> Option<Instant> {
+        Some(Instant(self.0.checked_add(duration)?))
+    }
+
+    /// Returns Some(t) where t is the time self - duration if t can be represented as Instant
+    /// (which means it’s inside the bounds of the underlying data structure), None otherwise.
+    pub fn checked_sub(&self, duration: Duration) -> Option<Instant> {
+        Some(Instant(self.0.checked_sub(duration)?))
+    }
+}
+
+impl Sub<Instant> for Instant {
+    type Output = Duration;
+
+    fn sub(self, other: Instant) -> Duration {
+        self.duration_since(other)
+    }
+}
+
+impl Add<Duration> for Instant {
+    type Output = Instant;
+
+    fn add(self, other: Duration) -> Instant {
+        self.checked_add(other)
+            .expect("overflow when adding duration to instant")
+    }
+}
+
+impl AddAssign<Duration> for Instant {
+    fn add_assign(&mut self, other: Duration) {
+        *self = *self + other;
+    }
+}
+
+impl Sub<Duration> for Instant {
+    type Output = Instant;
+
+    fn sub(self, other: Duration) -> Instant {
+        self.checked_sub(other)
+            .expect("overflow when subtracting duration from instant")
+    }
+}
+
+impl SubAssign<Duration> for Instant {
+    fn sub_assign(&mut self, other: Duration) {
+        *self = *self - other;
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::{Duration, Instant};
+    use std::thread;
+
+    #[test]
+    fn test_instant() {
+        let start = Instant::now();
+
+        let ms = 100;
+        thread::sleep(Duration::from_millis(ms));
+
+        let elapsed = start.elapsed();
+        assert!(elapsed.as_millis() as u64 >= ms - 1);
+        assert!(elapsed.as_millis() as u64 <= ms + 1);
+    }
+}
--- a/src/libos/crates/async-rt/src/time/mod.rs
+++ b/src/libos/crates/async-rt/src/time/mod.rs
@@ -1,0 +1,6 @@
+pub(crate) use self::entry::{TimerEntry, TimerFutureEntry};
+pub use self::instant::{Instant, DURATION_ZERO};
+
+mod entry;
+mod instant;
+mod wheel;
--- a/src/libos/crates/async-rt/src/time/wheel.rs
+++ b/src/libos/crates/async-rt/src/time/wheel.rs
@@ -1,0 +1,103 @@
+use super::entry::TimerWheelEntry;
+use super::Instant;
+use crate::prelude::*;
+use spin::mutex::MutexGuard;
+
+use hierarchical_hash_wheel_timer::wheels::quad_wheel::QuadWheelWithOverflow;
+
+lazy_static! {
+    pub static ref TIMER_WHEEL: TimerWheel = {
+        use crate::executor::EXECUTOR;
+        use crate::sched::yield_;
+        use crate::task::Task;
+
+        let wheel = TimerWheel::new();
+
+        // TODO: Don't always run this task when no useful tasks to run and no timeouts to expire.
+        let task = Arc::new(Task::new(async {
+            loop {
+                TIMER_WHEEL.try_make_progress();
+                yield_().await;
+            }
+        }));
+        EXECUTOR.accept_task(task);
+
+        wheel
+    };
+}
+
+/// The interface of hierarchical timer wheel. The resolution of time is 1ms.
+pub struct TimerWheel {
+    // hierarchical timer wheel.
+    wheel: Mutex<QuadWheelWithOverflow<TimerWheelEntry>>,
+    // current ticks, one tick means 1ms.
+    ticks: AtomicU64,
+    // start time of the timerwheel.
+    start: Instant,
+}
+
+impl TimerWheel {
+    pub fn new() -> Self {
+        Self {
+            wheel: Mutex::new(QuadWheelWithOverflow::default()),
+            ticks: AtomicU64::new(0),
+            start: Instant::now(),
+        }
+    }
+
+    /// Insert a new timer entry to the wheel and return the ticks when the entry is inserted.
+    pub fn insert_entry(&self, entry: TimerWheelEntry, timeout: Duration) -> u64 {
+        let mut guard = self.wheel.lock();
+
+        // Try to make progress to assure that the wheel is up to date.
+        let entries = self.make_progress(&mut guard);
+
+        // The minimum resolution of QuadWheelWithOverflow is 1 ms.
+        // If the timeout less than 1 ms, wait 1ms instead.
+        let timeout = core::cmp::max(timeout, Duration::MILLISECOND);
+        guard.insert_with_delay(entry, timeout).unwrap();
+        let insert_ticks = self.latest_ticks();
+
+        drop(guard);
+        Self::fire(entries);
+
+        insert_ticks
+    }
+
+    /// Try to move the timerwheel forward and fire expired timers.
+    pub fn try_make_progress(&self) {
+        if let Some(mut wheel_guard) = self.wheel.try_lock() {
+            let entries = self.make_progress(&mut wheel_guard);
+            drop(wheel_guard);
+            Self::fire(entries);
+        }
+    }
+
+    /// Try to move the timerwheel forward and return all expired timers.
+    ///
+    /// The returned timers must be fired by `fire()`!
+    fn make_progress(
+        &self,
+        wheel_guard: &mut MutexGuard<QuadWheelWithOverflow<TimerWheelEntry>>,
+    ) -> Vec<TimerWheelEntry> {
+        let elapsed = self.start.elapsed().as_millis() as u64;
+        // calculate the step that we need move forward, in most times, it should be 0 or 1.
+        let diff = elapsed - self.latest_ticks();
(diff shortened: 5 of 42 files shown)


src/libos/crates/async-rt/src/time/timer.rs
```diff
@@ -0,0 +1,200 @@
...
+use crate::time::Instant;
+
+use hierarchical_hash_wheel_timer::wheels::quad_wheel::QuadWheelWithOverflow;
+
+lazy_static! {
+    pub static ref TIMER_WHEEL: TimerWheel = TimerWheel::new();
+}
+
+/// The interface of hierarchical timer wheel.
+/// The resolution of time is 1ms.
```
tatetian, line 13: There is a concept called timer slack in Linux.

> The timer slack is the length of time by which the kernel may delay the wake-up of certain system calls that block with a timeout. Permitting this delay allows the kernel to coalesce wake-up events, thus possibly reducing the number of system wake-ups and saving power... The "current" timer slack is used by the kernel to group timer expirations for the calling thread that are close to one another; as a consequence, timer expirations for the thread may be up to the specified number of nanoseconds late. Grouping timer expirations can help reduce system power consumption by minimizing CPU wake-ups... The timer slack values of init (PID 1), the ancestor of all processes, are 50,000 nanoseconds (50 microseconds).

So what is our value of timer slack? Is it 1ms?

src/libos/crates/async-rt/src/time/timer.rs
```diff
@@ -0,0 +1,200 @@
...
+        }
+    }
+
+    /// Move the timerwheel forward.
+    fn make_process(&self, wheel_guard: &mut MutexGuard<QuadWheelWithOverflow<TimerWheelEntry>>) {
+        // We use the untrusted clock. If we get non-decreasing time unfortunately, use 0 instead.
+        let elapsed = self
+            .start
+            .checked_elapsed()
+            .map_or(0, |e| e.as_millis() as u64);
```
tatetian, line 62: If the clock is decreasing, which is an unexpected and non-recoverable error, I think we should simply panic, instead of hiding  such abnormal circumstances.

src/libos/crates/async-rt/src/time/timeout.rs
```diff
@@ -0,0 +1,54 @@
...
+            **this.timeout = this.timer.remain_duration();
+            return Poll::Ready(Ok(output));
+        }
+
+        // Second, we poll the timer.
+        // If the timer is ready, the timeout has expired, return Ready.
+        // else return Pending to wait next poll.
+        match this.timer.poll(cx) {
+            Poll::Ready(_) => {
+                **this.timeout = Duration::new(0, 0);
```
tatetian, line 48: Use `Duration::ZERO` instead.

tatetian (commented):
To make code review more smooth, I am hoping that future submitted commits can be more _atomic_. Please break down big commit like one into more digestible parts in the future.

src/libos/crates/async-rt/src/wait/waiter.rs
```diff
@@ -40,6 +44,17 @@ impl Waiter {
         self.inner.wait()
     }
 
+    #[cfg(feature = "timeout")]
+    pub fn wait_timeout<'a, T: BorrowMut<Duration>>(
+        &'a self,
+        timeout: Option<&'a mut T>,
+    ) -> WrapperFuture<WaitFuture<'_>, Timeout<WaitFuture<'_>>> {
```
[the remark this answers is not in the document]
ShuochengWang: I agree that using `impl Future<Output=Result<()>> + '_` as the return type.

About another issue,  if we combine these two functions, this will cause the user program to change as feature changes. I'm not sure if this is a good idea.

src/libos/crates/async-rt/src/wait/waiter.rs
```diff
@@ -155,3 +170,67 @@ impl<'a> Future for WaitFuture<'a> {
...
+#[pin_project]
+pub struct WrapperFuture<F1, F2> {
+    #[pin]
+    future: UnionFuture<F1, F2>,
+}
+
+/// A pinned enum to support union format.
+/// Since it's constructed by the pin_project crate, the real name of this enum is `UnionFutureProj`.
+#[pin_project(project = UnionFutureProj)]
+pub enum UnionFuture<F1, F2> {
```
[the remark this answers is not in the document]
ShuochengWang: The doc of `select!`: Polls multiple futures and streams simultaneously.
What we want is a union future type, which is only one future, instead of multiple futures.
ps, I got the idea of the `UnionFuture` enum from the async_std crate.

src/libos/crates/vdso-time/Cargo.toml
```diff
@@ -7,11 +7,11 @@ edition = "2018"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [features]
-default = ["libc"]
+default = []
 sgx = ["sgx_types", "sgx_tstd", "sgx_libc"]
 
 [dependencies]
-libc = { version = "0.2", optional = true }
+libc = { version = "0.2" }
```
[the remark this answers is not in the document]
ShuochengWang: It's a problem caused by no_std `async-rt`.
If `libc` is optional, we need specify `libc` by `default` or feature named like `std`.
```
default = ["libc"]
# or
std = ["libc"]
```
In `async-rt` crate, we can enable vdso-time in sgx by `vdso-time/sgx` feature and disable vdso-time `default` feature.  However, it would cause some trouble when `cargo t`, since there are no libc in vdso-time in this case.

src/libos/crates/async-rt/src/time/instant.rs
```diff
@@ -0,0 +1,220 @@
+use crate::prelude::*;
+use core::ops::{Add, AddAssign, Sub, SubAssign};
+use vdso_time::Vdso;
+
+const NSEC_PER_SEC: u64 = 1_000_000_000;
```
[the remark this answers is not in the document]
ShuochengWang: I agree.

src/libos/crates/async-rt/src/time/instant.rs
```diff
@@ -0,0 +1,220 @@
...
+lazy_static! {
+    static ref VDSO: Vdso = Vdso::new().unwrap();
+}
+
+fn now(clockid: vdso_time::clockid_t) -> Timespec {
+    let mut tp = vdso_time::timespec {
+        tv_sec: 0,
+        tv_nsec: 0,
+    };
+    VDSO.clock_gettime(clockid, &mut tp).unwrap();
```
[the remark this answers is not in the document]
ShuochengWang: This is a problem left over by history, since the vdso-time crate is serving for libos's linux API syscalls.
I agree to refactor the API to `Duration`.

src/libos/crates/async-rt/src/time/instant.rs
```diff
@@ -0,0 +1,220 @@
+use crate::prelude::*;
+use core::ops::{Add, AddAssign, Sub, SubAssign};
+use vdso_time::Vdso;
+
+const NSEC_PER_SEC: u64 = 1_000_000_000;
+
+/// A measurement of a monotonically nondecreasing clock. Opaque and useful only with Duration.
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Ord, PartialOrd)]
+pub struct Instant {
+    t: Timespec,
```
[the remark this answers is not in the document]
ShuochengWang: I agree

src/libos/crates/async-rt/src/time/mod.rs
```diff
@@ -0,0 +1,6 @@
+pub mod instant;
+pub mod timeout;
+pub mod timer;
```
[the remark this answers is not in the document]
ShuochengWang: I agree. There are some problems about pub / private in this commit, I will deal with these details.

src/libos/crates/async-rt/src/time/timeout.rs
```diff
@@ -0,0 +1,54 @@
...
+///
+/// Since it's a future, it doesn't execute unless you `.await` or `poll`.
+#[pin_project]
+pub struct Timeout<'a, F> {
+    #[pin]
+    future: F,
+    #[pin]
+    timer: TimerEntry,
+
+    timeout: &'a mut Duration,
```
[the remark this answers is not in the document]
ShuochengWang: I agree to deal with the special case.
Since we can not use `Option<TimerEntry>` because of `pin` problem, we need to create `TimeEntry`. Fortunately, creating `TimeEntry` isn't costly. 
1. we can record this special case in a field of `Timeout`, like `wait_forever`,  just poll future `F` in `poll`
2. we can deal with this special case in `TimerEntry`, just return `Pending` in `poll`.

src/libos/crates/async-rt/src/time/timer.rs
```diff
@@ -0,0 +1,200 @@
+pub use core::task::Waker as RawWaker;
```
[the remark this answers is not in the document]
ShuochengWang: it should be private, and no rename.

src/libos/crates/async-rt/src/time/instant.rs
```diff
@@ -0,0 +1,220 @@
...
+}
+
+impl Instant {
+    /// Returns an instant corresponding to “now”.
+    ///
+    /// Rust std says that the clock of linux + x86 environment is real-monotonic.
+    /// When sgx feature enabled, we are in linux + x86 env, the clock should be real-monotonic.
+    /// However, the clock in sgx env is untrusted, we can not guarantee that clock is real-monotonic.
+    /// If the host attacks us by provide wrong time, althrough we can record the `LAST_NOW`
+    /// to ensure the instant is monotonically nondecreasing, the time we get may not make sense at all.
```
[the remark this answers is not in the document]
ShuochengWang: If add `LAST_NOW`, we will need use `SgxThreadMutex`. rust sgx sdk uses `LAST_NOW` scheme

src/libos/crates/async-rt/src/wait/macros.rs
```diff
@@ -62,6 +62,68 @@ macro_rules! waiter_loop {
     }};
 }
 
+/// A waiter queue-coordinated loop with timeout.
+///
+/// Just like a waiter_loop, except that this loop will be forced to exit when the timeout expired.
+/// The the timeout expired, the loop will break, user should deal with the timeout case by themselves.
+#[cfg(feature = "timeout")]
+#[macro_export]
+macro_rules! waiter_timeout_loop {
```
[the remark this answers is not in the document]
ShuochengWang: I am not familar with macro, it's a very good combination,

src/libos/crates/async-rt/src/wait/mod.rs
```diff
@@ -60,6 +75,21 @@ mod tests {
                 });
             }
 
+            #[cfg(feature = "timeout")]
+            pub async fn consume_with_timeout(&self, timeout: Option<&mut Duration>) -> Option<T> {
```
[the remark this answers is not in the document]
ShuochengWang: I agree.

src/libos/crates/async-rt/src/time/timer.rs
ShuochengWang: Well, we don't have the timer slack concept. By since the 1ms resolution, the smallest time unit is 1ms. It mean that if wait for 1 ms and 50 microseconds, we will wait for 2 ms.
The 1ms is the restriction from hierarchical timer wheel crate, since the resolution supported by this crate is 1ms. 
Notes that, 1 ms = 1000 microseconds,  which equal to 20 timer slacks.
If the 1ms is too coarse for ngo, we can impl a timer wheel with higher resolution.

src/libos/crates/async-rt/src/time/timer.rs
ShuochengWang: If we use `LAST_NOW`, the clock will be non-decreasing.

src/libos/crates/async-rt/src/time/timeout.rs
ShuochengWang: I tried it before, `ZERO` is not supported by our env's rust version.

src/libos/crates/async-rt/src/wait/waiter.rs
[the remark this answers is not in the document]
tatetian: > I agree that using `impl Future<Output=Result<()>> + '_` as the return type.

I didn't see the changes. You have a second thought?

> About another issue, if we combine these two functions, this will cause the user program to change as feature changes. I'm not sure if this is a good idea.

I agree. This is a valid point. Let's keep the two methods.

src/libos/crates/async-rt/src/wait/mod.rs
```diff
@@ -100,4 +130,125 @@ mod tests {
...
+        }
+
+        #[test]
+        fn wait_none() {
+            crate::task::block_on(async {
+                let waiter = Waiter::new();
+                let waker = waiter.waker();
+
+                let join_handle = crate::task::spawn(async move {
+                    waiter.wait_timeout::<Duration>(None).await.unwrap();
```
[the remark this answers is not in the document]
ShuochengWang: Well, it's really ugly... 
I only use it here to test, fortunately, user will not use API as this. The user will use `.wait()` or .`wait_timeout(timeout)`

src/libos/crates/async-rt/src/wait/macros.rs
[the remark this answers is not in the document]
ShuochengWang: We can combine two macro, but we can not  use `$crate::waiter_loop!($waiter_queue, None, $loop_body);` to simplify it. The simplified version will might cause the type of return value mismatched if the waiter_loop is in the ending of function.

src/libos/crates/async-rt/src/wait/waiter.rs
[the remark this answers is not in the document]
ShuochengWang: Remove the `timeout` feature and make `timeout` as necessary instead of optional.
And we still keep two methods.

src/libos/crates/async-rt/src/wait/waiter.rs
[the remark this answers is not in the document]
ShuochengWang: Use `select_biased` macro
``` Rust
select_biased! {
                    _ = self.inner.wait().fuse() => { ... }
                    _ = TimerFutureEntry::new(&timer_entry).fuse() => { ... }
};
```

src/libos/crates/async-rt/src/wait/waiter.rs
```diff
@@ -155,3 +170,67 @@ impl<'a> Future for WaitFuture<'a> {
...
+            *raw_waker = None;
+            self.waiter.set_state(WaiterState::Idle);
+        }
+    }
+}
+
+/// A wrapper to unify the future type.
+/// This wrapper can support two future types and store only one future.
+#[pin_project]
+pub struct WrapperFuture<F1, F2> {
```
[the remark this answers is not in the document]
ShuochengWang: Removed.

src/libos/crates/vdso-time/Cargo.toml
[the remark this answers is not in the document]
ShuochengWang: vdso-time crate cargo.toml:
```
[features]
default = ["std"]
std = ["libc"]
sgx = ["sgx_types", "sgx_tstd", "sgx_libc"]
```

async-rt cargo.toml
```
[features]
default = ["vdso-time/std"]
auto_run = []
sgx = ["sgx_tstd", "flume/sgx", "vdso-time/sgx", "hierarchical_hash_wheel_timer/sgx"]
```

src/libos/crates/async-rt/src/time/instant.rs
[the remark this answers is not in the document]
ShuochengWang: Resolved.

src/libos/crates/async-rt/src/time/timeout.rs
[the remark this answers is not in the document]
ShuochengWang: The special case is resolved in new implementations.

src/libos/crates/async-rt/src/time/instant.rs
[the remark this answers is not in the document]
ShuochengWang: `LAST_NOW` is added.

src/libos/crates/async-rt/src/executor.rs
```diff
@@ -73,6 +73,17 @@ impl Executor {
...
         assert!(run_queue_id < self.parallelism);
         let run_queue = &self.run_queues[run_queue_id as usize];
+
+        if run_queue_id == 0 {
+            let task = Arc::new(Task::new(async {
+                loop {
+                    crate::time::TIMER_WHEEL.make_process();
+                    crate::sched::yield_().await;
+                }
+            }));
```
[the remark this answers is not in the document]
tatetian: Rename `TimerWheel::make_process` to `make_progress`.

src/libos/crates/async-rt/src/time/entry.rs
```diff
@@ -0,0 +1,165 @@
...
+            TimerState::Cancelled(_) => panic!("Can not cancel twice"),
+        }
+    }
+
+    /// Get the remained duration.
+    pub fn remained_duration(&self) -> Duration {
+        match self.state {
+            TimerState::Init => self.timeout,
+            TimerState::Started(start_tick) => {
+                let used_tick = TIMER_WHEEL.latest_tick() - start_tick;
```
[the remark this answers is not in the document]
tatetian: Consider epochs.

src/libos/crates/async-rt/src/wait/macros.rs
```diff
@@ -13,9 +13,21 @@ use super::{Waiter, WaiterQueue};
...
+/// The timeout argument should be an Option of Duration, e.g., `Option<Duration>`
+/// `Option<&mut Duration>`
 #[macro_export]
 macro_rules! waiter_loop {
     ($waiter_queue:expr, $loop_body:block) => {{
+        // The reason why we don't just use `$crate::waiter_loop!($waiter_queue, None, $loop_body)`:
+        // If we use this simplified syntax, we can not deal with the return value in some case.
+        // Since we need `break` when reached timeout, `break` the loop may return `()`
+        // if the waiter_loop is in the ending of function. If user want to return in loop_body,
+        // the compiler will report error: the type of return value is mismatched.
```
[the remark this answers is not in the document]
ShuochengWang: Failed when try to use `break res`.
The result type is hard to match the return type in loop body and the function return type.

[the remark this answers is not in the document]
ShuochengWang:
```Rust
            pub async fn consume(&self, timeout: Option<&mut Duration>) -> Result<T> {
                crate::waiter_loop!(&self.consumers, timeout, {
                        return Ok(item);
                    }
                })
            }
```
If we use `break res`, there will be three results: 
1. the err result when timeout, Err(&str)
2. the ok result in the loop body, Ok(T)
3. the result in the function return value, Result<T>

To use the waiter_loop in libos, we need to use `errno` crate in async_rt. But there are still problems: user will get trouble when the user want a custom error, which happens often.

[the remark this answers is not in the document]
ShuochengWang: My thoughts:
1. just use `break`:
```Rust
    waiter_loop!(process.sig_waiters(), timeout, {
        if let Some(signal) = dequeue_pending_signal(&interest, &thread, &process) {
            let siginfo = signal.to_info();
            return Ok(siginfo);
        }
    });
    return_errno!(EAGAIN, "no interesting, pending signal");
```
2.use a `timeout_body`:
```Rust
    waiter_loop!(process.sig_waiters(), timeout, {
        if let Some(signal) = dequeue_pending_signal(&interest, &thread, &process) {
            let siginfo = signal.to_info();
            return Ok(siginfo);
        }
    }, {
        return_errno!(EAGAIN, "no interesting, pending signal");
    });
```

And in both cases, we should split the implementation of waiter_loop(timeout) and wait_loop() to simplify the use of `wait_loop()`

src/libos/crates/vdso-time/Cargo.toml
```diff
@@ -7,18 +7,25 @@ edition = "2018"
...
+log = "0.4"
 sgx_types = { path = "../../../../deps/rust-sgx-sdk/sgx_types", optional = true }
 sgx_tstd = { path = "../../../../deps/rust-sgx-sdk/sgx_tstd", optional = true, features = ["backtrace"] }
 sgx_libc = { path = "../../../../deps/rust-sgx-sdk/sgx_libc", optional = true }
+sgx_trts = { path = "../../../../deps/rust-sgx-sdk/sgx_trts", optional = true }
 
 [dev-dependencies]
 criterion = "0.3"
 lazy_static = { version = "1.4.0", features = ["spin_no_std"] }
+ctor = "0.1"
```
tatetian, line 28: I didn't see code using this crate.

src/libos/crates/vdso-time/ocalls/vdso-time-ocalls.c
```diff
@@ -5,39 +5,27 @@
...
-    } else { *coarse_resolution = 0; }
-
     struct utsname buf;
     int ret = uname(&buf);
+    // uname should always succeed here, since uname only fails when buf is not invalid.
     if (ret != 0) { return -1; }
 
     int buf_rel_len = strlen(buf.release);
     int len = buf_rel_len > release_len ? release_len : buf_rel_len;
     memcpy(release, buf.release, len);
```
tatetian, line 20: As a rule of thumb, variable-length C strings should always be null terminated. All fields in utsname are null-terminated.

src/libos/crates/vdso-time/src/lib.rs
```diff
@@ -9,289 +7,284 @@ extern crate sgx_types;
...
+            4 => ClockId::CLOCK_MONOTONIC_RAW,
+            5 => ClockId::CLOCK_REALTIME_COARSE,
+            6 => ClockId::CLOCK_MONOTONIC_COARSE,
+            7 => ClockId::CLOCK_BOOTTIME,
+            _ => return_errno!(EINVAL, "Unsupported clockid"),
+        })
+    }
+}
+
+/// An abstraction of Linux vDSO, privide get time interface through Linux vDSO.
```
tatetian, line 63: The sentence is not correct in grammar.

src/libos/crates/vdso-time/src/lib.rs
```diff
@@ -9,289 +7,284 @@ extern crate sgx_types;
...
-            let hres_resolution = if ret == 0 { Some(tp.tv_nsec) } else { None };
-            let ret = unsafe { libc::clock_getres(CLOCK_REALTIME_COARSE, &mut tp as *mut _) };
-            let coarse_resolution = if ret == 0 { Some(tp.tv_nsec) } else { None };
-
-            let mut utsname: libc::utsname = unsafe { std::mem::zeroed() };
-            let ret = unsafe { libc::uname(&mut utsname as *mut _) };
-            if ret != 0 {
-                return Err(());
-            }
+    /// Try to Create a new Vdso by libc or SGX OCALL.
```
tatetian, line 73: "Create" --> "create"

src/libos/crates/vdso-time/src/lib.rs
```diff
@@ -9,289 +7,284 @@ extern crate sgx_types;
...
-
-        // Compare the results of Linux and vdso
-        // to check whether vdso can support the clockid correctly.
-        for &clockid in &clockids {
-            if vdso.support_clocks[clockid as usize] {
-                let mut tp = timespec {
-                    tv_sec: 0,
-                    tv_nsec: 0,
+    /// Compare the results of Linux syscall and vdso to check whether vdso can support the clockid correctly.
+    fn get_checked_vdso(vdso: Self) -> Result<Self> {
```
tatetian, line 187: Usually, a getter method is simple and do not change the state of an object. This method is complex and consumes `self`. So a `get_xxx` name is not a good fit for this method. How about `validate_accuracy` or `check_accuracy`?

src/libos/crates/vdso-time/src/lib.rs
```diff
@@ -9,289 +7,284 @@ extern crate sgx_types;
...
+            return_errno!(EINVAL, "Vdso get utsname failed");
+        }
+        let release = utsname.release;
 
-            (vdso_addr, hres_resolution, coarse_resolution, tss, release)
-        };
+        Self::match_kernel_version(vdso_addr, &release)
+    }
 
+    fn get_checked_vdso_addr(vdso_addr: u64) -> Result<u64> {
```
tatetian, line 136: Similar to the other comment, we don't want call a method `get_xx` if it is not a getter actually. Renaming this method to `check_vdso_addr` seems to be a good idea to me. And you don't need to consume the subject of a check method (like you do for `get_checked_vdso_addr` or `get_checked_vdso`). A signature of `fn check_vdso_addr(vdso_addr: &u64) -> Result<()>` looks fine to me.

src/libos/crates/vdso-time/src/lib.rs
```diff
@@ -9,289 +7,284 @@ extern crate sgx_types;
...
         if vdso_addr == 0 {
-            return Err(());
+            return_errno!(EFAULT, "Vdso vdso_addr is 0")
         }
+        let vdso_data_max_size = 4 * PAGE_SIZE;
```
tatetian, line 140: This should be a const.

tatetian (commented):
Here are my comments for commit "[vdso-time] Refactor code and add fallback". The commit looks good. There are only some minor issues. Thanks.

src/libos/crates/async-rt/src/time/instant.rs
```diff
@@ -0,0 +1,125 @@
...
+    /// When sgx feature enabled, we are in linux + x86 env, the clock should be real-monotonic.
+    /// However, the clock in sgx env is untrusted, we can not guarantee that clock is real-monotonic.
+    /// To ensure the instant is monotonically nondecreasing, We keep a global "latest now" instance
+    /// which is returned instead of what the OS says if the OS goes backwards.
+    pub fn now() -> Self {
+        let os_now = Instant(clock_gettime(ClockId::CLOCK_MONOTONIC).unwrap());
+
+        let mut last_now = LAST_NOW.lock();
+        let now = core::cmp::max(*last_now, os_now);
+        *last_now = now;
```
tatetian, line 28: This implementation is good. A straightforward (and buggy) implementation would check `os_now` against `LAST_NOW`, which could result in false-positive failures in some race conditions involving concurrent updates to `LAST_NOW`.

src/libos/crates/vdso-time/src/lib.rs
ShuochengWang: /// An abstraction of Linux vDSO provides the clock and time interface through Linux vDSO.

src/libos/crates/vdso-time/src/lib.rs
ShuochengWang: changed to `validate_accuracy`

src/libos/crates/async-rt/src/time/wheel.rs
```diff
@@ -0,0 +1,103 @@
...
+    pub fn new() -> Self {
+        Self {
+            wheel: Mutex::new(QuadWheelWithOverflow::default()),
+            ticks: AtomicU64::new(0),
+            start: Instant::now(),
+        }
+    }
+
+    /// Insert a new timer entry to the wheel.
+    /// Return the latest tick of the timerwheel, or Err() if the duration is too short (less than 1ms).
```
tatetian, line 49: Do you think it is a good idea to round up the timeout of `entry` to 1ms so that the insertion can always succeed? Instead of documenting about all possible errors of an API, it is more preferable to eliminate errors in the first place. Infallible APIs are most user friendly. We can even guarantee this property of minimal timeout value by changing the constructor of `TimerEntry`.

By the way, it seems that you have checked the property in a later phase.

```rust
                // The minimum resolution of QuadWheelWithOverflow is 1 ms.
                // If the timeout less than 1 ms, wait 1ms instead.
                let timeout = core::cmp::max(shared.timeout, Duration::MILLISECOND);
```

src/libos/crates/vdso-time/src/lib.rs
ShuochengWang: changed to `fn check_vdso_addr(vdso_addr: u64) -> Result<()>`

tatetian (commented):
Here are my comments for commit "[async-rt] Add time module for timeout". The commit looks fine. Only one minor issue.

src/libos/crates/async-rt/src/wait/macros.rs
```diff
@@ -6,21 +6,39 @@ use super::{Waiter, WaiterQueue};
...
-        use $crate::wait::{AutoWaiter, Waiter, WaiterQueue};
+        $crate::waiter_loop!($waiter_queue, None::<core::time::Duration>, $loop_body)
+    }};
+    ($waiter_queue:expr, $timeout:expr, $loop_body:block) => {{
+        use $crate::wait::{AutoWaiter, WaiterQueue};
 
         let waiter_queue: &WaiterQueue = $waiter_queue;
         let mut auto_waiter = AutoWaiter::new(waiter_queue);
         let mut loop_count: u32 = 0;
+        let mut timeout = $timeout.map_or(None, |t| Some(t));
```
tatetian, line 41: What is the purpose of `map_or` here? The `map_or` does not seem to have any effect.

src/libos/crates/async-rt/src/wait/macros.rs
```diff
@@ -6,21 +6,39 @@ use super::{Waiter, WaiterQueue};
 /// is used to avoid busy loop. The user code of the loop body can use
 /// `break`, `continue`, or `return` as it would do in a normal loop.
 ///
+/// However, there are restrictions on the return value, the user code
+/// must return a `Result` when it `break` or `return`. If the user code
+/// doesn't use `break` and the waiter_loop isn't the return value, the user
+/// code need to use explicit annotation to tell compiler the type of `Result`.
+/// ```
+///
```
tatetian, line 14: Some sample code is missing here.

src/libos/crates/vdso-time/Cargo.toml
ShuochengWang: It's commit error, this code is using by next commit

src/libos/src/process/do_futex.rs
```diff
@@ -144,16 +144,15 @@ pub async fn futex_wait_bitset(
...
-
-    /* TODO: handle errror
-    if let Err(e) = res {
+    let mut timeout = timeout.map_or(None, |t| Some(t));
+    // Wait until we wake it up or reach timeout.
+    if let Err(_) = waiter.wait_timeout(timeout.as_mut()).await {
         let (_, futex_bucket_ref) = FUTEX_BUCKETS.get_bucket(futex_item.key);
         let mut futex_bucket = futex_bucket_ref.lock().unwrap();
         futex_bucket.dequeue_item(&futex_item);
-        return_errno!(e.errno(), "futex wait timeout or interrupted");
```
tatetian, line 155: Actually, I think you can keep the original code of returning `e.errno()` as it would save you from changing it again when `Waiter::wait_timeout` can be interrupted.

src/libos/src/process/do_futex.rs
```diff
@@ -293,13 +298,13 @@ impl FutexBucket {
     }
 
     // TODO: this is an O(N) operation. Try to make it more efficient
-    /*pub fn dequeue_item(&mut self, futex_item: &FutexItem) -> Option<FutexItem> {
+    pub fn dequeue_item(&mut self, futex_item: &FutexItem) -> Option<FutexItem> {
```
tatetian, line 301: I am confused. If `dequeue_item` is disabled before, how could the original code compile?

src/libos/crates/async-rt/src/time/wheel.rs
ShuochengWang: Since I checked the property before insertion, the insertion will always succeed.
I agree to move the check to insertion function, hence the insertion will always succeed and doesn't need the comments anymore.

src/libos/src/signal/do_sigtimedwait.rs
```diff
@@ -26,13 +22,15 @@ pub async fn do_sigtimedwait(interest: SigSet, timeout: Option<&Duration>) -> Re
...
+    let mut timeout = timeout.cloned();
     // Loop until we find a pending signal or reach timeout
-    waiter_loop!(process.sig_waiters(), {
+    waiter_loop!(process.sig_waiters(), timeout, {
         if let Some(signal) = dequeue_pending_signal(&interest, &thread, &process) {
             let siginfo = signal.to_info();
             return Ok(siginfo);
         }
-    })
+    }) as Result<siginfo_t>;
```
tatetian, line 32: It is a bit of verbose to write `as Result<_>` and then `return_errno`. I think it can be simplified this way.

```rust
waiter_loop!(process.sig_waiters(), timeout, {
    //...
}).map_err(|e| errno!(EAGAIN, "no interesting, pending signal"))
```

In the future, when `wait` may return `EINTR`, then the code should be written like this.

```rust
waiter_loop!(process.sig_waiters(), timeout, {
    //...
}).map_err(|e| {
    if e.errno() == ETIMEDOUT { 
        errno!(EAGAIN, "no interesting, pending signal")
    } else {
        e
    }
})
```

The above version highlights the fact that `waiter_loop` always returns a `Result<_>` and the user can deal with it (instead of ignoring it).

tatetian (commented):
Here is my comments for commit "[libos] Support timeout in sigtimedwait syscall".

tatetian:
Overall, I am really satisfied with the quality of this PR. Good job and thanks a lot for the efforts in improving code!

src/libos/crates/async-rt/src/time/wheel.rs
ShuochengWang:
```Rust
    /// Insert a new timer entry to the wheel and return the ticks when the entry is inserted.
    pub fn insert_entry(&self, entry: TimerWheelEntry, timeout: Duration) -> u64 {
        let mut guard = self.wheel.lock();

        // Try to make progress to assure that the wheel is up to date.
        let entries = self.make_progress(&mut guard);

        // The minimum resolution of QuadWheelWithOverflow is 1 ms.
        // If the timeout less than 1 ms, wait 1ms instead.
        let timeout = core::cmp::max(timeout, Duration::MILLISECOND);
        guard.insert_with_delay(entry, timeout).unwrap();
        let insert_ticks = self.latest_ticks();

        drop(guard);
        Self::fire(entries);

        insert_ticks
    }
```

src/libos/crates/async-rt/src/wait/macros.rs
ShuochengWang: the purpose is to create a `mut timeout` if the timeout is not `mut`

src/libos/src/process/do_futex.rs
ShuochengWang: But there is a problem: we return `ETIME` in `wait_timeout`, but here we need to return `ETIMEDOUT`.
Should we return `ETIMEDOUT` in `wait_timeout`?

src/libos/src/process/do_futex.rs
ShuochengWang: it's because that the code using it is commented on before.
```
/*
        let (_, futex_bucket_ref) = FUTEX_BUCKETS.get_bucket(futex_item.key);
        let mut futex_bucket = futex_bucket_ref.lock().unwrap();
        futex_bucket.dequeue_item(&futex_item);
*/
```

src/libos/src/signal/do_sigtimedwait.rs
ShuochengWang: According to linux doc, the errno should ETIMEDOUT

> ETIMEDOUT
>               The operation in futex_op employed the timeout specified
>               in timeout, and the timeout expired before the operation
>               completed.

> EAGAIN (FUTEX_WAIT, FUTEX_WAIT_BITSET, FUTEX_WAIT_REQUEUE_PI) The
>               value pointed to by uaddr was not equal to the expected
>               value val at the time of the call.

```
waiter_loop!(process.sig_waiters(), timeout, {
    //...
}).map_err(|e| {
    if e.errno() == ETIME { 
        errno!(ETIMEDOUT, "no interesting, pending signal")
    } else {
        e
    }
})
```

src/libos/crates/vdso-time/ocalls/vdso-time-ocalls.c
ShuochengWang:
```
    strncpy(release, buf.release, release_len);
    release[release_len - 1] = '\0';
```

src/libos/src/process/do_futex.rs
ShuochengWang: ETIME  Timer expired (POSIX.1 (XSI STREAMS option)).

              (POSIX.1 says "STREAM ioctl(2) timeout".)

       ETIMEDOUT
              Connection timed out (POSIX.1-2001).

src/libos/crates/async-rt/src/time/wheel.rs
tatetian: Just to be sure. The timeout value being rounded up is not visible to users, right? For example, the return value of `remained_duration ` method.

src/libos/src/process/do_futex.rs
tatetian: I don't even know there is an errno code named `ETIME`! Could you give an example system call that returns `ETIME`?

Still, my judgement (you can challenge) favors `ETIMEDOUT`. The reason is that from the user perspective, he/she calls `Waiter::wait`, not create a timer directly. The `Waiter::wait` method is most like futex wait, which would return `ETIMEDOUT` if timeout.

PR #33, opened 2021-08-09
[test] Fix race condition bug in make test.

ShuochengWang:
- Enable `make test`.
- Fix race condition bug in `make test`.
     - bug details: make test use `occlum start` to test, which use occlum server mode. Whenever the server receives a task, the server will spawn a thread to check whether the task is completed and returned. This thread will init a futex_value with `-1`, and then `futex_wait(0)`. When the task is completed, libos will set the futex_value with task's retval and `futex_wake`. And in most times, the retval is `0`. If the thread started very slow and `futex_wait` is later than `futex_wake`, and the retval is `0`, will cause the thread sleep forever.
     - The `futex_wait` value should equal to init value, and should not in the [0, 0xff00] range, since the retval might be an interger in [0, 0xff00].

### DIFF
--- a/src/exec/src/server.rs
+++ b/src/exec/src/server.rs
@@ -211,7 +211,9 @@ impl OcclumExec for OcclumExecImpl {
             thread::spawn(move || {
                 loop {
                     debug!("waiting:");
-                    match exit_status.wait(0) {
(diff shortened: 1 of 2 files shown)


tatetian (approved):
LGTM

PR #35, opened 2021-08-17
Update toolchain and enable github action

ShuochengWang:
- Fix compiling error of external ocalls
- Update toolchain and rust sgx sdk
- Fix sefs compiling error temporarily
    - TODO: update sefs crate and update libos/fs code
- Rename .github.disable folder to enable github action
- Disable glibc test, use temp docker image
    - TODO: enable glibc test and use stantard docker image


.github/workflows/main.yml
```diff
@@ -20,7 +20,10 @@ jobs:
       run: echo "OCCLUM_VERSION=$(grep 'Version =' src/pal/include/occlum_version.h | awk '{print $4}')" >> $GITHUB_ENV;
 
     - name: Create container with ubuntu image
-      run: docker run -itd --name=ubuntu-test -v $GITHUB_WORKSPACE:/root/occlum occlum/occlum:${{ env.OCCLUM_VERSION }}-ubuntu18.04
+      # run: docker run -itd --name=ubuntu-test -v $GITHUB_WORKSPACE:/root/occlum occlum/occlum:${{ env.OCCLUM_VERSION }}-ubuntu18.04
+      # TODO: This is a temporary dockerhub repo for ngo, need update dockerhub repo later.
+      # The new docker image should install libclang-dev
+      run: docker run -itd --name=ubuntu-test -v $GITHUB_WORKSPACE:/root/occlum shuochengwang/ngo:0.23.7-ubuntu18.04
```
tatetian, line 26: It is ok to let everyone uses this Docker image of yours. But I didn't see any updates to Ubuntu Dockerfile. Without these, how can anyone know exactly what changes that NGO require for its Docker image? And how can anyone to continue updates the Docker image without knowing the new Dockerfile?

tools/docker/Dockerfile.ubuntu18.04
```diff
@@ -74,7 +74,7 @@ RUN git clone -b sgx_2.11_for_occlum https://github.com/occlum/linux-sgx && \
 
 # Install Rust
 ENV PATH="/root/.cargo/bin:$PATH"
```
tatetian, line 76: In the updates to Git submodule, the branch `sgx_2.13_for_occlum` of Rust SGX SDK is selected. But in the Dockerfile, the branch of Intel SGX SDK is `sgx_2.11_for_occlum`. Is this inconsistency intentional or should be fixed? There is no guarantee that Rust SGX SDK 2.x is compatible with Intel SGX SDK 2.y, if x is not equal to y.

ShuochengWang: I haven't update the dockerfile, since the dockerfile is meaningless now, I think we should update dockerfile according to latest occlum.

ps. The temporary docker image is created by the dockerfile of occlum 0.23.7 (with clang installed)

PR #40, opened 2021-09-07
Support timeout for poll / epoll && Fix bugs of poll / epoll

ShuochengWang:


src/libos/src/poll/do_epoll/epoll_file.rs
```diff
@@ -227,6 +228,10 @@ impl EpollFile {
                     entry.reset_ready();
                 }
             }
+
+            if ready.len() == 0 || count_events >= max_events {
+                break;
+            }
```
tatetian, line 234: You are right that the original loop is missing the ending condition. But the condition is not `ready.len() == 0 || count_events >= max_events`. The reason is that `ready.len()` may never reach zero since some ready entries---those that are neither edge-triggered or one-shot---will be kept in the entry list (see line 223). So whether the ready list is empty or not does not matter when deciding when to stop.

I think this is the right way to terminate the loop.

```rust
let pop_count = (max_events - count_events).min(pop_quota);
if pop_count == 0 {
    return;
}
```

This is basically saying in natural language that "we want to return as many results as possible. so if we cannot pop any more candidates, then it is time to give up by exiting the loop."

tatetian (commented):
Here are my comments.

src/libos/crates/async-io/src/event/event_counter.rs
```diff
@@ -28,6 +31,15 @@ impl EventCounter {
         .unwrap()
     }
 
+    pub async fn read_timeout(&self, timeout: &mut Option<&mut Duration>) -> Result<u64> {
```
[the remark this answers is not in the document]
ShuochengWang: Current code using mut ref

```Rust
// do_poll()
    loop {
        ...
        // Return if the timeout expires.
        if let Err(_) = poller.wait_timeout(&mut timeout).await {
            return Ok(0);
        }
    }

// Poller
pub async fn wait_timeout(&self, timeout: &mut Option<&mut Duration>) { ... }

// eventcounter
pub async fn read_timeout(&self, timeout: &mut Option<&mut Duration>) -> Result<u64> {
    waiter_loop!(&self.waiters, timeout, { ... })
}
```

If we simplified it to `Option<&mut Duration>`, we need to use `mut timeout: Option<&mut Duration>` as args of `poller.wait_timeout()` and `read_timeout`, and we need to create a new `option` every time in the `do_poll()` loop

src/libos/crates/async-rt/src/wait/mod.rs
```diff
@@ -162,6 +162,62 @@ mod tests {
         });
     }
 
+    #[test]
+    fn test_timeout_args_type() {
```
tatetian, line 166: Breaking this test into two tests would make the code more readable.

PR #41, opened 2021-09-08
Support getpeername and getsockname syscalls

ShuochengWang:


ShuochengWang:
getsockname for different states:
- Init
    - bind: `addr` is `Some`, return it
    - not bind: `addr` is `None`, get addr from host but don't store this addr (This addr will be invalid in later state, we should not store it)
- Connect: get addr from host but don't store this addr 
- Connected: return `addr` if it's `Some`, else get addr from host and store this addr
- Listen: return `addr` if it's `Some`, else get addr from host and store this addr

src/libos/crates/host-socket/src/stream/mod.rs
```diff
@@ -185,6 +185,47 @@ impl<A: Addr, R: Runtime> StreamSocket<A, R> {
...
+                |addr| Ok(addr),
+            ),
+            State::Listen(listener_stream) => listener_stream.common().addr().map_or_else(
+                || {
+                    let addr = listener_stream.common().get_addr_from_host()?;
+                    listener_stream.common().set_addr(&addr);
+                    Ok(addr)
+                },
+                |addr| Ok(addr),
+            ),
```
tatetian, line 215: The cases of connected and listening sockets can be combined to avoid code duplication and make the fact that their logic is the same explicit.

src/libos/crates/host-socket/src/stream/mod.rs
```diff
@@ -185,6 +185,47 @@ impl<A: Addr, R: Runtime> StreamSocket<A, R> {
...
+        let state = self.state.read().unwrap();
+        match &*state {
+            State::Init(init_stream) => init_stream.common().addr().map_or_else(
+                || init_stream.common().get_addr_from_host(),
+                |addr| Ok(addr),
+            ),
+            State::Connect(connecting_stream) => connecting_stream.common().addr().map_or_else(
+                || connecting_stream.common().get_addr_from_host(),
+                |addr| Ok(addr),
+            ),
```
tatetian, line 199: Could you shred some lights on why init or connecting sockets should call `get_addr_from_host`? For unix stream sockets, whose addresses are unnamed by default? (I don't if this should be the right behavior. Need some testing to confirm.) If so, I think the more correct logic is to `stream.common().addr().unwrap_or_default()`.

ShuochengWang: > If the socket is not bound to an address, the call returns with the family set, and **the rest of the structure is set to 0**.

> The getsockname function is often used to find the port assigned to a socket after the socket has been implicitly bound to a port. For example, an application can call connect without previously calling bind. In this case, the connect function completes the binding necessary by assigning a port to the socket. This assignment can be found with a call to getsockname.

> A UNIX domain stream socket that is not bound to a path name by using the bind function is **unnamed**. Similarly, the two UNIX domain sockets that are created by using the socketpair function are **unnamed**. When the address of an unnamed socket is returned, a NULL value is returned for the sun_path field.

tatetian (approved):

PR #44, opened 2021-09-23
Support getsockopt and setsockopt

ShuochengWang:


src/libos/crates/host-socket/src/common/sockopt.rs
```diff
@@ -0,0 +1,280 @@
...
+        use libc::getsockopt as do_getsockopt;
+        use libc::setsockopt as do_setsockopt;
+    }
+}
+
+use crate::prelude::*;
+
+const SOL_SOCKET: i32 = 1;
+
+pub fn new_getsockopt_cmd(
```
tatetian, line 20: A function taking raw pointers can hardly be safe (and this function is definitely not). Either mark this function as unsafe and document how to use it safely or refactor the type of `optval` to `&mut [u8]`.

src/libos/crates/host-socket/src/common/sockopt.rs
```diff
@@ -0,0 +1,280 @@
...
+            input.optval,
+            input.optlen,
+        )?;
+        self.set_output(());
+        Ok(())
+    }
+}
+
+impl_ioctl_cmd! {
+    pub struct SoAcceptConnCmd<Input=(), Output=i32> {}
```
tatetian, line 157: I suggest replace the `So` prefix (it basically contains zero info as the type belongs to a module named `sockopt`) with `Get` (it helps to differentiate between get or set commands) in the four commands of `SoAcceptConnCmd`, `SoDomainCmd`, `SoPeerNameCmd`, and `SoTypeCmd`.

src/libos/src/net/syscalls.rs
```diff
@@ -289,3 +348,18 @@ fn copy_sock_addr_to_user(
     unsafe { *dst_addr_len = src_addr_len as u32 };
     Ok(())
 }
+
+fn put_to_user_unchecked<T>(
+    data: T,
+    user_val_ptr: *mut c_void,
+    user_len_ptr: *mut libc::socklen_t,
+) -> Result<()> {
```
tatetian, line 356: This function is unsafe

src/libos/crates/host-socket/src/lib.rs
```diff
@@ -13,9 +13,13 @@ extern crate sgx_tstd as std;
...
 #[macro_use]
 mod prelude;
+mod common;
 mod runtime;
 mod stream;
 mod util;
 
+pub use self::common::sockopt::{
+    new_getsockopt_cmd, new_setsockopt_cmd, SoAcceptConnCmd, SoDomainCmd, SoPeerNameCmd, SoTypeCmd,
+};
```
tatetian, line 23: I think sockopt-related features do not qualify as the top-level APIs. Making the `sockopt` module public and exposing them beneath the module is better.

src/libos/crates/host-socket/src/common/sockopt.rs
```diff
@@ -0,0 +1,280 @@
...
+        SockOptName::SO_PEERNAME => Box::new(SoPeerNameCmd::new(())),
+        SockOptName::SO_TYPE => Box::new(SoTypeCmd::new(())),
+        SockOptName::SO_CNX_ADVICE => return_errno!(ENOPROTOOPT, "it's a write-only option"),
+        _ => Box::new(GetSockOptRawCmd::new(GetSockOptRawInput::new(
+            level, optname, optval, optlen,
+        ))),
+    })
+}
+
+pub fn new_setsockopt_cmd(
```
tatetian, line 46: If you think more about it, the `new_getsockopt_cmd` and `new_setsockopt_cmd ` methods belongs more to the LibOS rather than the `host-socket` crate since they serve as a C-to-Rust or Linux-to-host_socket adapter.

src/libos/crates/host-socket/src/common/sockopt.rs
```diff
@@ -0,0 +1,280 @@
...
+        | SockOptName::SO_SNDLOWAT
+        | SockOptName::SO_PEERSEC
+        | SockOptName::SO_PROTOCOL
+        | SockOptName::SO_MEMINFO
+        | SockOptName::SO_INCOMING_NAPI_ID
+        | SockOptName::SO_COOKIE
+        | SockOptName::SO_PEERGROUPS => return_errno!(ENOPROTOOPT, "it's a read-only option"),
+        _ => Box::new(SetSockOptRawCmd::new(SetSockOptRawInput::new(
+            level, optname, optval, optlen,
+        ))),
```
tatetian, line 76: I suggest combining `SetSockOptRawCmd` and `SetSockOptRawInput` into one. Do not use the `impl_ioctl_cmd` macro; simply `impl IoctlCmd for SetSockOptRawCmd`.

src/libos/crates/host-socket/src/common/sockopt.rs
```diff
@@ -0,0 +1,280 @@
...
+    })
+}
+
+#[derive(Debug)]
+pub struct GetSockOptRawInput {
+    level: c_int,
+    optname: c_int,
+    optval: *mut c_void,
```
tatetian, line 84: Avoid using raw pointers casually. Try this.

```rust
pub struct GetSockOptRaw<'a> {
    level: i32,
    optname: i32,
    optval: &'a mut [u8],
}
```

tatetian (commented):
The commit mostly looks good. But I think it has some room for improvement. See my comments below. Thanks.

src/libos/src/net/syscalls.rs
```diff
@@ -200,6 +205,76 @@ pub async fn do_getsockname(
...
+        fd, level, optname, optval, optlen
+    );
+
+    let file_ref = current!().file(fd as FileDesc)?;
+    let socket_file = file_ref
+        .as_socket_file()
+        .ok_or_else(|| errno!(EINVAL, "not a socket"))?;
+
+    from_user::check_mut_ptr(optlen)?;
+    from_user::check_mut_array(optval, unsafe { *optlen } as _)?;
```
tatetian, line 250: In my opinion, the best practice when handling user raw pointers is to check pointers and then convert them to their Rust counterparts.

```rust
let optlen_mut = {
    from_user::check_mut_ptr(optlen)?;
    &mut *optlen
};
let optlen = *optlen_mut;
let optval_mut = {
    from_user::check_mut_array(optval, optlen as usize)?;
    slice::from_raw_parts_mut(optval as *mut u8, optlen)
};
```

After this point, you can write safe Rust code on `optlen_mut`, `optlen`, and `optval_mut`. This way, the function `put_to_user_unchecked`, `new_getsockopt_cmd`, `new_setsockopt_cmd` won't be unsafe anymore.

I see that you added the safety sections for these unsafe functions. But no explanations can be better than---let's say---no need to explain at all (when unsafe functions do not exist at all).

**Do not write comments to explain bad code. Write good code instead.**

src/libos/crates/host-socket/src/common/sockopt.rs
```diff
@@ -0,0 +1,197 @@
...
+        }
+    }
+
+    pub fn getsockopt(&mut self, fd: HostFd) -> Result<()> {
+        getsockopt_by_host(fd, self.level, self.optname, self.optval, self.optlen)?;
+        Ok(())
+    }
+}
+
+impl IoctlCmd for GetSockOptRawCmd<'static> {}
```
tatetian, line 41: The LibOS can only construct valid `GetSockOptRawCmd<'a>` for `'a != 'static`, but only `GetSockOptRawCmd<'static>` implements `IoctlCmd`. This means that there must be some highly unsafe code converting `GetSockOptRawCmd<'a>` to `GetSockOptRawCmd<'static>`. And yes, as expected, in a later commit, the unsafe conversion happens in `new_setsockopt_cmd` (a big unsafe function, which itself is a bad smell); it happens in a very stealth way (there is no explicit conversion). 

Converting `T<'a>` to `T<'static>` basically means workaround Rust life times and the borrow checker (then why use Rust at all?). There must be a very, very strong reason to do so. And I don't think that is our situation.

Let's discuss offline for a better solution.

src/libos/crates/host-socket/src/common/sockopt.rs
```diff
@@ -0,0 +1,197 @@
...
+impl_ioctl_cmd! {
+    pub struct GetDomainCmd<Input=(), Output=i32> {}
+}
+
+pub struct AddrStorage(pub (libc::sockaddr_storage, usize));
+impl std::fmt::Debug for AddrStorage {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_tuple("AddrStorage")
+            .field(&"sockaddr_storage")
+            .field(&(self.0).1)
```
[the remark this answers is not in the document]
ShuochengWang: It's because of a bug of cargo fmt.
If use `&self.0.1`, cargo fmt will crash

src/libos/crates/host-socket/src/common/sockopt.rs
```diff
@@ -0,0 +1,197 @@
...
+    optname: i32,
+    optval: &mut [u8],
+    optlen: &mut u32,
+) -> Result<()> {
+    try_libc!(do_getsockopt(
+        fd as _,
+        level as _,
+        optname as _,
+        optval.as_mut_ptr() as _,
+        optlen as *mut u32
```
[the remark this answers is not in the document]
ShuochengWang: I check the rust sdk code, rust sdk doesn't guarantee that.

tatetian (commented):
Now everything looks great. Except one thing. See my inline comments.

src/libos/crates/host-socket/src/sockopt/mod.rs
```diff
@@ -0,0 +1,97 @@
...
+    SO_TIMESTAMPNS_NEW = 64,         // i32, bool, 0 or 1, cmsg-related
+    SO_TIMESTAMPING_NEW = 65,        // i32, bool, 0 or 1, cmsg-related
+    SO_RCVTIMEO_NEW = 66,            // struct timeval
+    SO_SNDTIMEO_NEW = 67,            // struct timeval
+    SO_DETACH_REUSEPORT_BPF = 68,    // BPF-related
+}
+
+pub trait GetOutputAsBytes {
+    fn get_output_as_bytes(&self) -> Option<&[u8]>;
+}
```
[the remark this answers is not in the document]
ShuochengWang: But there is another problem: `do_getsockopt` only know the name of these sockoptcmds (e.g., GetDomainCmd), libos should know the specific output type to implement the trait
```Rust
impl GetOutputAsBytes for GetDomainCmd {
    fn get_output_as_bytes(&self) -> Option<&[u8]> {
        self.output().map(|val_ref| unsafe {
            std::slice::from_raw_parts(val_ref as *const _ as *const u8, std::mem::size_of::<i32>())
        })
    }
}
```

tatetian (approved):

PR #45, opened 2021-09-26
Enable glibc test

ShuochengWang:


src/libos/src/process/do_clone.rs
```diff
@@ -30,13 +30,16 @@ pub async fn do_clone(
...
+    // a relatively safe address in case the user program uses %fs before
+    // initializing fs base address.
     let init_cpu_state = CpuContext {
         gp_regs: GpRegs {
             rsp: user_rsp as _,
             rip: thread_entry as _,
             ..Default::default()
         },
-        fs_base: new_tls.unwrap_or(0) as _,
+        fs_base: new_tls.unwrap_or(current!().vm().get_stack_limit()) as _,
```
guzongmin, line 42: The stack information in vm is used by the first thread, If the current thread is not the first one, the implementation maybe wrong.

guzongmin (approved):
LGTM

PR #46, opened 2021-09-28
Support nonblocking mode

ShuochengWang:
- Fix a bug of io-uring-callback: `iohandle.await` can not be waken up
- Support nonblocking mode
- Support status flags (o_nonblock)


tatetian (approved):
Good job. I am looking forward for more high-quality commits that require zero revision 😄 Thanks for the contribution!

PR #51, opened 2021-10-14
Support datagram socket (UDP) and select syscall

ghost:


src/libos/crates/host-socket/src/stream/mod.rs
```diff
@@ -308,6 +308,15 @@ impl<A: Addr, R: Runtime> StreamSocket<A, R> {
...
     }
 
+    pub fn cancel_requests(&self) {
+        let state = self.state.read().unwrap();
+        match &*state {
+            State::Listen(listener_stream) => listener_stream.cancel_requests(),
+            State::Connected(connected_stream) => connected_stream.cancel_requests(),
+            _ => {}
+        }
+    }
```
tatetian, line 318: A `cancel_requests` method should not be made public. A public method should not be named `cancel_requests`. This is because having async, ongoing io_uring requests are entirely implementation details. We should not leak such information to the users. Futhermore, I think the `Drop` method should be implemented for `StreamSocket`, instead of `HostSocket`. Remember crates like `host_socket` should work fine on its own without the LibOS. Implementing `Drop` for LibOS's `HostSocket` cannot solve the resource release problem of `StreamSocket`.

The same argument applies to `DatagramSocket`. But I am not sure that implementing `Drop` for `DatagramSocket` can guarantee cancelling requests. The io_uring requests for `StreamSocket` does not hold any `Arc` of `StreamSocket`. But is this the case for `DatagramSocket`?

Lastly, I think there is a race condition where after `cancel_requests` is called, a completion callback function of a io_uring request is invoked, which in turn submits another io_uring requests. This race condition should be fixed to ensure that resources are released in all circumstances.

src/libos/src/poll/do_epoll/epoll_file.rs
```diff
@@ -167,7 +167,11 @@ impl EpollFile {
             // If no ready entries for now, wait for them
             if poller.is_none() {
                 poller = Some(Poller::new());
-                self.pollee.poll(Events::IN, poller.as_mut());
+            }
+
+            let events = self.pollee.poll(Events::IN, poller.as_mut());
+            if !events.is_empty() {
+                continue;
```
tatetian, line 174: I didn't get it. I think the changed code is equivalent to the original version.

PR #53, opened 2021-10-19
Use goblin as elf parser and only load segments on demand

jessehui:
Libos now Only read in headers and needed segments. Also reduce a memory copy.


guzongmin (approved):
LGTM

PR #91, opened 2021-11-05
fix system hang issue

guzongmin:
1. Support interruptible timer
    + in some case (nano_sleep, futex, ....), the waiter is not able to be interrupted to handle signal 
2. Support async wake
    + if the two threads sync with futex are scheduled on same vcpu, the system would hang

### DIFF
--- a/src/libos/crates/async-rt/src/task/mod.rs
+++ b/src/libos/crates/async-rt/src/task/mod.rs
@@ -3,6 +3,7 @@ use self::task::TaskBuilder;
 use crate::executor::EXECUTOR;
 use crate::prelude::*;
 use crate::sched::SchedPriority;
+use crate::wait::Signal;
 
 pub use self::id::TaskId;
 pub use self::join::JoinHandle;
@@ -17,8 +18,11 @@ mod join;
 mod locals;
 mod task;
 
-pub fn spawn<T: Send + 'static>(future: impl Future<Output = T> + 'static + Send) -> JoinHandle<T> {
-    SpawnOptions::new(future).spawn()
+pub fn spawn<T: Send + 'static>(
+    future: impl Future<Output = T> + 'static + Send,
+    signal: Option<Arc<Signal>>,
+) -> JoinHandle<T> {
+    SpawnOptions::new(future).signal(signal).spawn()
 }
 
 pub fn block_on<T: Send + 'static>(future: impl Future<Output = T> + 'static + Send) -> T {
@@ -65,6 +69,7 @@ fn init_runner_threads() {
 pub struct SpawnOptions<T> {
     raw_future: Option<BoxFuture<'static, T>>,
     priority: SchedPriority,
+    signal: Option<Arc<Signal>>,
 }
 
 impl<T: Send + 'static> SpawnOptions<T> {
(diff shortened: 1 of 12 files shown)


tatetian (commented):
In our prior dicussions, I was positive about the way interrupts are supported as shown in this PR. But after examing the full code, I don't think the solution is satisfying for two reasons.

First and foremost, the solution is not entirely correct. Consider the race conditin in a simple setup where a thread A is sending a signal to thread B, when the latter one is going to sleep. The flow of execution can be boiled down to this.

* Thread A:
```rust
let waiter_queue = self.sig_waiters();
waiter_queue.wake_all();
```
* Thread B:
```rust
let mut auto_waiter = AutoWaiter::new(waiter_queue);
let mut waiter = auto_waiter.waiter();
waiter.wait().await;
```

If `waiter.wait()` is executed before `waiter_queue.wake_all()`, then we can wake thread B succesfully. But what if the ordering is reversed? 

(I will pause here to leave you some time to think)

The answer is, unfortunately, that the waiter will not be woken up. This is because `WaiterQueue` is a queue in essense: its `wake_all` method can only wake up the waiters that have already been enqueued.

The second problem is about complexity. The proposal makes `Waiter::wait` 10X more complex than the original code. Interrupts are relatively rare events; but synchronization (by `Waiter::wait`) is frequent. It is highly undesirable to make it common code path slow just to make a rare case possible. I wouldn't say that the extra complexity brings a huge overhead. The overhead is acceptable. But the code just doesn't look quite right to me.

```rust
let task = current::get().clone();
let waiter_queue = task.irq_waiters();

if let Some(waiter_queue) = waiter_queue {
    let mut auto_waiter = AutoWaiter::new(waiter_queue);
    let mut waiter = auto_waiter.waiter();

    select_biased! {
        _ = self.inner.wait().fuse() => {}
        _ = waiter.inner.wait().fuse() => {
            return_errno!(EINTR, "the waiter interrupted");
        }
    }
} else {
    self.inner.wait().await;
}
```

On top of the two problems above, I notice another issue about `waiter_loop`. This macro calls `waiter::wait` or `waiter::wait_timeout` internally. The interrupt support will adds a new possibility of return value (`Err(EINTR)`), which breaks the assumption of the current `waiter_loop` implementation. In general, we should examine all invocations of `Waiter::wait` and see if adding interrupts breaks assumptions or not.

In conclusion, I think adding the interrupt support is more complex than it first appears. I am also considering alternative approaches, but have not made my mind yet.

tatetian (commented):
I appreciate the new proposal. It fixes thhe race condition that I mentioned in previous comments. But still it is not ready to be accepted.

Let's put aside other "soft" standards (like whether the API of `asyc_rt::wait::Signal` is considered good or not). The new proposal is still buggy. The interrupt mechanism works by connecting `SigQueues`, `Task` and `Waiter::wait` through the new primitive of `Signal` or (`SigCount`). In case you didn't notice, there are actually two kinds of `SigQueues`, one is per-process, the other is per-thread. The per-thread `SigQueues` are used to deliver thread-targeted signals, which should also be able to interrupt. But it won't; the current implementation is only able to interrupt for process-targeted signals.

PR #118, opened 2021-11-23
[net] Add trusted unix domain socket implementation

jessehui:


src/libos/crates/async-io/src/util/channel.rs
```diff
@@ -158,9 +181,83 @@ impl Producer {
...
+        }
+    }
+
+    fn try_writev(&self, bufs: &[&[u8]]) -> Result<usize> {
+        let mut total_nbytes = 0;
+        for buf in bufs {
+            let ret = self.write(buf);
+            if let Ok(nbytes) = ret {
+                total_nbytes += nbytes;
+            }
```
guzongmin, line 249: Should we return the error here?

src/libos/crates/async-io/src/util/channel.rs
```diff
@@ -262,9 +357,82 @@ impl Consumer {
...
+        }
+    }
+
+    fn try_readv(&self, bufs: &mut [&mut [u8]]) -> Result<usize> {
+        let mut total_nbytes = 0;
+        for buf in bufs {
+            let ret = self.read(buf);
+            if let Ok(nbytes) = ret {
+                total_nbytes += nbytes;
+            }
```
guzongmin, line 424: return the error

src/libos/crates/async-io/src/util/channel.rs
jessehui: Refine the logic here. Thanks for your review.

src/libos/crates/async-io/src/util/channel.rs
jessehui: ditto.

jessehui:
This PR has merged with #135

PR #125, opened 2021-12-01
[libos] Fix epoll file's handling with closed files

### LINKED ISSUE #48: Fix the bug of epoll: EEXIST error when two different files with the same fd
Bug:
1.  first add one file with fd `x` to epoll interest
2. close this file
3. create a new file, which will reuse the fd `x` since `FileTable` will reuse fd
4. add the new file to epoll interest
5. epoll will report `EEXIST` error!

Might Related Todos:
- // TODO: add fd to FileHandle?
- // TODO: make FileTable a struct of internal mutability to avoid deadlock.
- // TODO: close host_fd

Solutions:
1. generate different fd in `FileTable`
2. Find a better design for: `do_close`, `drop`, `observer`, the close of socket file (which is `Arc<...>`)

liqinggd:
Fixes #48

### DIFF
--- a/src/libos/src/fs/file_table.rs
+++ b/src/libos/src/fs/file_table.rs
@@ -1,4 +1,6 @@
 use super::*;
+use crate::poll::EpollFile;
+use std::sync::Weak;
 
 pub type FileDesc = u32;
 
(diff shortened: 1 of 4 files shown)


guzongmin (approved):
LGTM

PR #126, opened 2021-12-01
Support interrupts

tatetian:
The PR consists of three main commits (among other minor commits):
* [async-rt] Add Task Interrupt ReQuests
* [libos] Force thread exit with interrupts
* [libos] Force signal delivery with interrupts

### DIFF
--- a/src/libos/src/process/thread/mod.rs
+++ b/src/libos/src/process/thread/mod.rs
@@ -1,4 +1,4 @@
-use async_rt::task::Task;
+use async_rt::task::{Task, Tirqs};
 use std::fmt;
 use std::ptr::NonNull;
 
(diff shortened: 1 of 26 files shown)


src/libos/src/process/thread/mod.rs
```diff
@@ -76,9 +76,33 @@ impl Thread {
...
+    /// This method also updates the TIRQ mask so that the two masks are kept in sync.
+    pub fn set_sig_mask(&self, mut new_sig_mask: SigSet) -> SigSet {
+        use crate::signal::{SIGKILL, SIGSTOP};
+        new_sig_mask -= SIGKILL;
+        new_sig_mask -= SIGSTOP;
+
+        let mut sig_mask = self.sig_mask.write().unwrap();
+        let old_sig_mask = *sig_mask;
+        *sig_mask = new_sig_mask;
+        drop(sig_mask);
```
guzongmin, line 99: Should we drop the sig_mask after  Tirqs::set_mask(new_sig_mask.to_c() as u64);?

tatetian: You see the following `debug_assert!`? This method is only supposed to be called in the context of the task of the thread. So this method won't be called concurrently. And so we don't need to keep the lock guard after `Tirqs::set_mask`.

src/libos/src/signal/do_sigtimedwait.rs
```diff
@@ -23,34 +24,22 @@ pub async fn do_sigtimedwait(interest: SigSet, timeout: Option<&Duration>) -> Re
...
+    let wait_err = wait_res.unwrap_err();
 
-fn dequeue_process_pending_signal(
-    process: &ProcessRef,
-    interest: &SigSet,
-) -> Option<Box<dyn Signal>> {
-    let blocked = !*interest;
-    process.sig_queues().write().unwrap().dequeue(&blocked)
-}
+    // If the waiter loop is interrupted, we try one last time
```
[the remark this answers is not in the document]
tatetian: After reconsidering the check, I think you are right. It is not necessary. I will remove it. Good point.

guzongmin (approved):
LGTM

PR #131, opened 2021-12-08
[demo] Add the FIO demo

liqinggd:


demos/fio/fio.yaml
```diff
@@ -0,0 +1,14 @@
+includes:
+  - base.yaml
+# sqlite
```
guzongmin, line 3: It is not a sqlite demo, so we should use "#fio".

demos/fio/run_fio_on_occlum.sh
```diff
@@ -0,0 +1,41 @@
...
+    echo "Please copy it into configs first"
+    exit 1
+fi
+
+# 1. Init Occlum instance
+rm -rf occlum_instance && occlum new occlum_instance
+cd occlum_instance
+new_json="$(jq '.resource_limits.max_num_of_threads = 128 |
+                .resource_limits.user_space_size = "320MB" |
+                .process.default_mmap_size = "256MB"' Occlum.json)" && \
```
guzongmin, line 30: .process.default_mmap_size = "256MB" is not required.

PR #151, opened 2021-12-20
Add timerfd

qzheng527:
Pass the LTP timerfd test and NGO timerfd test.


tatetian (commented):
I like the main idea of how timerfd is implemented in this PR. For reads on `TimerFd`, you check the timeout counter (`get_alarm_count`). If no good, you wait for the remaining of the time time to elapse. For poll on `TimerFd`, you check the timeout counter. If no good, you start a task to notify the pollee when the remaining time elapsed. This approach is simple and should be efficient enough.

However, to be 100% correct, the implementation needs two major improvements.

The first one is about epoll support. Epoll files are registered as the observers of a file-like object through `register_observer`. Observers get notified of I/O events when `Pollee::add_events` invoked on their target pollees. So, the implementation of a type of files (say `TimerFd`) needs to call `add_events` _whenever_ a type of I/O event turns active (e.g., `TimerFd` transits from unreadable to readable). The current implementation fails to do so.

The second one is that `TimerFd` is not thread safe.

```rust
pub struct TimerFd {
    clockid: ClockId,
    flags: Atomic<TimerCreationFlags>,
    fire_time: Atomic<Duration>,
    cur_interval: Atomic<Duration>,
    alarm_cnt: Atomic<usize>,
    last_alarm_cnt: Atomic<usize>,
    status: Atomic<TimerFdStatus>,
    pollee: Pollee,
}
```

The fields of `TimerFd` must be kept in sync regardless of any concurrent access, e.g., `read`, `poll`, `time`, `set_time`, etc. `AtomicXxx` and `Atomic<T>` are efficient, but they are less powerful than `Mutex` and `RwLock`. Using `Mutex` would be just the first step. You may notice some more subtle concurrency issues if you start thinking hard about the concurrent behaviors of the code.

The above are the two major problems. I will add inline comments for minor issues.

src/libos/src/time/timer_fd.rs
```diff
@@ -0,0 +1,297 @@
...
+#[derive(Clone, Copy, Debug, PartialEq)]
+#[repr(u8)]
+enum TimerFdStatus {
+    ONE_SHOT = 0,
+    LOOP = 1,
+    STOP = 2,
+}
+
+#[derive(Debug)]
+pub struct TimerFd {
```
tatetian, line 21: To be consistent with other file types (`EpollFile`, `EventFile`, `SocketFile`, etc.), this type should be named `TimerFile`.

src/libos/src/time/timer_fd.rs
```diff
@@ -0,0 +1,297 @@
...
+
+    pub fn poll(&self, mask: Events, poller: Option<&mut Poller>) -> Events {
+        let (count, left) = self.get_alarm_cnt();
+        if count > 0 {
+            Events::IN
+        } else {
+            self.pollee.del_events(Events::IN);
+            //let (tx, rx): (Sender<(Pollee, Duration)>, Receiver<(Pollee, Duration)>)= mpsc::channel();
+            let (tx, rx): (Sender<(Pollee, Duration)>, Receiver<(Pollee, Duration)>) =
+                flume::unbounded();
```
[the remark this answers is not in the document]
qzheng527: Nice, it is much simpler ,will do the update.

src/libos/src/time/mod.rs
```diff
@@ -203,3 +207,25 @@ impl TimeProvider for OcclumTimeProvider {
...
 }
+
+// For Timerfd
+#[repr(C)]
+#[derive(Debug, Default, Copy, Clone)]
+#[allow(non_camel_case_types)]
+pub struct itimerspec_t {
+    it_interval: timespec_t,
+    it_value: timespec_t,
+}
```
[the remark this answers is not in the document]
qzheng527: "clear cut" makes sense, will do the update.

src/libos/src/time/timer_fd.rs
qzheng527: Sounds reasonal, will do the change.

qzheng527:
> I like the main idea of how timerfd is implemented in this PR. For reads on `TimerFd`, you check the timeout counter (`get_alarm_count`). If no good, you wait for the remaining of the time time to elapse. For poll on `TimerFd`, you check the timeout counter. If no good, you start a task to notify the pollee when the remaining time elapsed. This approach is simple and should be efficient enough.
> 
> However, to be 100% correct, the implementation needs two major improvements.
> 
> The first one is about epoll support. Epoll files are registered as the observers of a file-like object through `register_observer`. Observers get notified of I/O events when `Pollee::add_events` invoked on their target pollees. So, the implementation of a type of files (say `TimerFd`) needs to call `add_events` _whenever_ a type of I/O event turns active (e.g., `TimerFd` transits from unreadable to readable). The current implementation fails to do so.
> 
> The second one is that `TimerFd` is not thread safe.
> 
> ```rust
> pub struct TimerFd {
>     clockid: ClockId,
>     flags: Atomic<TimerCreationFlags>,
>     fire_time: Atomic<Duration>,
>     cur_interval: Atomic<Duration>,
>     alarm_cnt: Atomic<usize>,
>     last_alarm_cnt: Atomic<usize>,
>     status: Atomic<TimerFdStatus>,
>     pollee: Pollee,
> }
> ```
> 
> The fields of `TimerFd` must be kept in sync regardless of any concurrent access, e.g., `read`, `poll`, `time`, `set_time`, etc. `AtomicXxx` and `Atomic<T>` are efficient, but they are less powerful than `Mutex` and `RwLock`. Using `Mutex` would be just the first step. You may notice some more subtle concurrency issues if you start thinking hard about the concurrent behaviors of the code.
> 
> The above are the two major problems. I will add inline comments for minor issues.

Thanks for the review comments, I will update the PR.

> I like the main idea of how timerfd is implemented in this PR. For reads on `TimerFd`, you check the timeout counter (`get_alarm_count`). If no good, you wait for the remaining of the time time to elapse. For poll on `TimerFd`, you check the timeout counter. If no good, you start a task to notify the pollee when the remaining time elapsed. This approach is simple and should be efficient enough.
> 
> However, to be 100% correct, the implementation needs two major improvements.
> 
> The first one is about epoll support. Epoll files are registered as the observers of a file-like object through `register_observer`. Observers get notified of I/O events when `Pollee::add_events` invoked on their target pollees. So, the implementation of a type of files (say `TimerFd`) needs to call `add_events` _whenever_ a type of I/O event turns active (e.g., `TimerFd` transits from unreadable to readable). The current implementation fails to do so.
> 
> The second one is that `TimerFd` is not thread safe.
> 
> ```rust
> pub struct TimerFd {
>     clockid: ClockId,
>     flags: Atomic<TimerCreationFlags>,
>     fire_time: Atomic<Duration>,
>     cur_interval: Atomic<Duration>,
>     alarm_cnt: Atomic<usize>,
>     last_alarm_cnt: Atomic<usize>,
>     status: Atomic<TimerFdStatus>,
>     pollee: Pollee,
> }
> ```
> 
> The fields of `TimerFd` must be kept in sync regardless of any concurrent access, e.g., `read`, `poll`, `time`, `set_time`, etc. `AtomicXxx` and `Atomic<T>` are efficient, but they are less powerful than `Mutex` and `RwLock`. Using `Mutex` would be just the first step. You may notice some more subtle concurrency issues if you start thinking hard about the concurrent behaviors of the code.
> 
> The above are the two major problems. I will add inline comments for minor issues.

Many thansk for all the review comments. 
I felt the concurrence would be a problem for my implemenation. I will consider your suggest "Mutex/RwLock" and update a new version for better concurrent support.
For the epoll, actually by my test and understanding, current implementation could support it already.
I will add a "epoll+timerfd" to the test in next udpated version.

qzheng527:
Two major improvements as below:
1. Use SgxMutexto protect the "TimerInner".
2. Once timer starts, spawn a task to do timer interval update (every trigger do `pollee.add_events(Events::IN)`) to satisfy epoll.

src/libos/src/time/timer_file.rs
```diff
@@ -0,0 +1,385 @@
...
+
+        if inner.status != TimerFdStatus::STOP {
+            let delta = match inner.fire_time.checked_sub(cur_time) {
+                Some(duration) => duration,
+                _ => Duration::new(0, 0),
+            };
+
+            if inner.status == TimerFdStatus::ONE_SHOT {
+                if delta.is_zero() == true {
+                    inner.alarm_cnt = 1;
```
guzongmin, line 199: Do not get the point why it should be 1? I think it could be 0 too.

src/libos/src/time/timer_file.rs
```diff
@@ -0,0 +1,385 @@
...
+            count = alarm.0;
+            left = alarm.1;
+
+            if count > 0 {
+                break;
+            } else if flags.contains(TimerCreationFlags::TFD_NONBLOCK) {
+                return_errno!(EAGAIN, "try again");
+            } else {
+                // block read, do wait for left time to rest cpu
+                if let Ok(_) = waiter.wait_timeout(Some(&mut left)).await {
```
guzongmin, line 263: Should it wait for the Events::IN event?

src/libos/src/time/timer_file.rs
qzheng527: `alarm_cnt=1` means there is one expired timer because the delta ( the delta of expected fire time and current time) is 0.
Anyway, I will try re-design the whole status/timer check to make it more readble.

qzheng527:
1. Use `Arc<Mutex<TimerInner>>` to support concurrence.
2. Update the states set and check.
3. Re-organize the code to use just one `wait_timeout` for the whole timerfd.

qzheng527:
Fix some corner case issue and add more comments.

src/libos/src/time/timer_file.rs
```diff
@@ -0,0 +1,394 @@
...
+            drop(inner);
+            self.start_timer_task(it_value);
+        }
+
+        return Ok(ret);
+    }
+
+    // The inner implementation of syscall timerfd_gettime.
+    pub fn time(&self) -> Result<TimerfileDurations> {
+        let mut ret_time = TimerfileDurations::default();
```
guzongmin, line 154: Maybe we could implement this function in inner. And TimerFile could simply invoke the  inner time function.

src/libos/src/time/timer_file.rs
```diff
@@ -0,0 +1,394 @@
...
+                        Err(e) => {
+                            if e.errno() == EINTR {
+                                // Error or Stopped by tirq
+                                break;
+                            } else if one_shot == true {
+                                // One-shot timer, no need loop
+                                trace!("timerfd one-shot triggerred");
+                                pollee.add_events(Events::IN);
+                                let mut inner = inner_c.lock().unwrap();
+                                inner.exp_cnt += 1;
```
guzongmin, line 232: Should we do inner.exp_cnt += 1 before add_events? I think the reader may read the count when getting the Events::IN

src/libos/src/time/timer_file.rs
```diff
@@ -0,0 +1,394 @@
...
+                                inner.status = TimerFdStatus::EXPIRE;
+                                inner.task_handle = None;
+                                break;
+                            } else {
+                                trace!("timerfd timer expired");
+                                // If it is not one-shot, the following timeout changes to interval.
+                                timeout = interval;
+                                pollee.add_events(Events::IN);
+                                let mut inner = inner_c.lock().unwrap();
+                                inner.exp_cnt += 1;
```
guzongmin, line 242: We could combine the common code:
      pollee.add_events(Events::IN);
      let mut inner = inner_c.lock().unwrap();
      inner.exp_cnt += 1;

src/libos/src/time/timer_file.rs
```diff
@@ -0,0 +1,394 @@
...
+            // Do not always hold it in the loop
+            drop(inner);
+
+            if count > 0 {
+                break;
+            } else if flags.contains(TimerCreationFlags::TFD_NONBLOCK) {
+                return_errno!(EAGAIN, "try again");
+            } else {
+                // Block wait, do yield
+                yield_().await;
```
guzongmin, line 290: It is not an efficient way. We should make sure the timer is triggered before check the exp_cnt.

src/libos/src/time/timer_file.rs
qzheng527: The inner is protected by SgxMutex. 
Put the function in inner implies the funtion will hold the lock all the time which is not efficient.

src/libos/src/time/timer_file.rs
qzheng527: Good catch, I will update it

src/libos/src/time/timer_file.rs
```diff
@@ -0,0 +1,401 @@
...
+        // Transfer the initial expired time to relative duration time.
+        // The relative time it_value will be used in wait_timeout.
+        let it_value = match exp_time.checked_sub(cur_time) {
+            Some(duration) => duration,
+            _ => Duration::new(0, 0),
+        };
+
+        inner.interval = new_value.it_interval;
+        inner.init_exp = exp_time;
+        inner.exp_cnt = 0;
```
guzongmin, line 135: Could we remove one of the number (  inner.exp_cnt , inner.last_exp_cnt)?

src/libos/src/time/timer_file.rs
```diff
@@ -0,0 +1,401 @@
...
+use atomic::{Atomic, Ordering};
+use std::sync::{Arc, SgxMutex as Mutex};
+use std::time::Duration;
+
+#[allow(non_camel_case_types)]
+#[derive(Clone, Copy, Debug, PartialEq)]
+#[repr(u8)]
+enum TimerFdStatus {
+    STOP = 0,
+    ONE_SHOT = 1,
```
guzongmin, line 14: ONE_SHOT and LOOP is not a status. They could be another enum

src/libos/src/time/timer_file.rs
```diff
@@ -0,0 +1,387 @@
...
+    }
+}
+
+impl TimerFile {
+    pub async fn read(&self, buf: &mut [u8]) -> Result<usize> {
+        if buf.len() < 8 {
+            return_errno!(EINVAL, "buffer is too small");
+        }
+
+        let count = self.wait_for_expire().await?;
```
guzongmin, line 314: The inner.exp_cnt may be increased after we get the count by let count = self.wait_for_expire().await, the line "inner.exp_cnt = 0" reset the count may miss some counts by mistake.

qzheng527: You are right. Do not reset it to 0 but reduce the expired count alread read could avoid the race.
```
        let mut inner = inner_c.lock().unwrap();
        // Reduce expired count alread read
        inner.exp_cnt = inner.exp_cnt - count;
```

src/libos/src/time/timer_file.rs
```diff
@@ -0,0 +1,367 @@
...
+}
+
+impl TimerFile {
+    pub async fn read(&self, buf: &mut [u8]) -> Result<usize> {
+        if buf.len() < 8 {
+            return_errno!(EINVAL, "buffer is too small");
+        }
+
+        let flags = self.flags.load(Ordering::Relaxed);
+        let mut count = 0;
```
guzongmin, line 284: Can we remove "let mut count = 0;"?

src/libos/src/time/timer_file.rs
```diff
@@ -0,0 +1,367 @@
...
+            let pollee = self.pollee.clone();
+            let exp_waiters = self.exp_waiters.clone();
+            let inner_c = self.inner.clone();
+
+            let expired_closure = move |inner_c: &Arc<Mutex<TimerInner>>, is_stop| {
+                let mut inner = inner_c.lock().unwrap();
+                inner.exp_cnt += 1;
+
+                if is_stop {
+                    inner.status = TimerFdStatus::EXPIRE;
```
guzongmin, line 208: Can we just set it to STOP? Seem the EXPIRE status is useless

guzongmin (approved):
LGTM

PR #153, opened 2021-12-20
Fix flink demo

jessehui:


src/libos/src/vm/mod.rs
```diff
@@ -59,6 +59,8 @@ running in the same Occlum instance can use dramatically different sizes of memo
 use super::*;
 use fs::{FileDesc, FileRef};
 use process::{Process, ProcessRef};
+use spin::rw_lock::{RwLockReadGuard, RwLockWriteGuard};
+use spin::RwLock;
```
guzongmin, line 63: If we already know the SGX_RwLock has a bug, we should replace all the RwLock with spin::RwLock as a workaround.

guzongmin (approved):
LGTM

PR #156, opened 2021-12-21
Add SGX 2 KSS support

qzheng527:


guzongmin (commented):
Can you please combine below two commits into one? 

Fix the C++ Link issue and Add the static lib 
Fix the compile error when including dcap_quote.h in pure C program

guzongmin (approved):
LGTM

PR #160, opened 2021-12-23
[libos] Fix compile warnings

qzheng527:
Most of them are used warnings.

### DIFF
--- a/src/libos/crates/async-io/src/event/poller.rs
+++ b/src/libos/crates/async-io/src/event/poller.rs
@@ -1,6 +1,6 @@
 use std::borrow::BorrowMut;
 use std::collections::HashMap;
-use std::hash::{Hash, Hasher};
+use std::hash::Hash;
 use std::sync::atomic::{AtomicU32, AtomicUsize, Ordering};
 use std::time::Duration;
 
@@ -219,7 +219,7 @@ impl Poller {
     /// Wait until there are any interesting events happen since last `wait`, or reach timeout.
     pub async fn wait_timeout<T: BorrowMut<Duration>>(
         &self,
-        mut timeout: Option<&mut T>,
+        timeout: Option<&mut T>,
     ) -> Result<()> {
         self.inner
             .event_counter
@@ -246,8 +246,8 @@ impl Drop for Poller {
         for weak_pollee in pollees.drain(..) {
             if let Some(pollee) = weak_pollee.upgrade() {
                 let mut pollers = pollee.pollers.lock();
-                let res = pollers.remove(&self_observer);
-                //assert!(res.is_some());
+                let _res = pollers.remove(&self_observer);
+                assert!(_res.is_some());
                 drop(pollers);
 
                 pollee.num_pollers.fetch_sub(1, Ordering::Relaxed);
--- a/src/libos/crates/async-io/src/file/file.rs
+++ b/src/libos/crates/async-io/src/file/file.rs
@@ -1,14 +1,10 @@
 use std::fmt::Debug;
-use std::marker::Unsize;
 use std::mem::transmute;
 use std::ops::CoerceUnsized;
-use std::ops::Deref;
 
-use futures::future::{self, BoxFuture};
-use futures::prelude::*;
 use inherit_methods_macro::inherit_methods;
 
-use crate::event::{Events, Observer, Pollee, Poller};
+use crate::event::{Events, Observer, Poller};
 use crate::file::{AccessMode, StatusFlags};
 use crate::fs::StatBuf;
 use crate::ioctl::IoctlCmd;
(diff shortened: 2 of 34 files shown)


src/libos/crates/async-io/src/file/file.rs
```diff
@@ -45,14 +41,17 @@ pub trait File: Debug + Sync + Send {
         Ok(0)
     }
 
+    #[allow(unused_variables)]
     fn poll(&self, mask: Events, poller: Option<&mut Poller>) -> Events {
```
liqinggd, line 45: I prefer to use this:
```
fn poll(&self, _mask: Events, _poller: Option<&mut Poller>) -> Events {
```

src/libos/crates/async-io/src/fs/stat_buf.rs
```diff
@@ -140,6 +140,7 @@ impl StatMode {
             FileType::BlockDevice => StatMode::BLOCK,
             FileType::Socket => StatMode::SOCKET,
             FileType::NamedPipe => StatMode::FIFO,
+            #[allow(unreachable_patterns)]
```
liqinggd, line 143: Why not just delete the `_ => StatMode::NULL`?

src/libos/crates/async-io/src/socket/addr/c_sock_addr.rs
```diff
@@ -42,7 +42,7 @@ impl CSockAddr for libc::sockaddr_in {
 
     fn c_addr(&self) -> &[u8] {
         // Note that libc::sockaddr_in contains a useless sin_zero field of 8 bytes. Do not use it.
-        let port_and_addr_len = size_of_val(&self.sin_port) + size_of_val(&self.sin_addr);
```
liqinggd, line 45: The code here seems to be useless, so why the author leaves the code here? If it is useless, we can just delete it.

src/libos/src/fs/syscalls.rs
```diff
@@ -627,9 +627,6 @@ pub async fn do_sendfile(
         from_user::check_mut_ptr(offset_ptr)?;
         Some(unsafe { offset_ptr.read() })
     };
-    if count < 0 {
```
liqinggd, line 630: cannot just delete the check. According to  the manual, the `count` argument can be a negative number,  you can change the type of `count` from `size_t` to `isize`

src/libos/crates/async-io/src/file/file.rs
qzheng527: Yes, it is another compiler suggest.

src/libos/crates/async-io/src/fs/stat_buf.rs
qzheng527: Keeping it here is because if more StatMode are added in future, there is no need add the default patterns again.
If no more StatMode will be added, delete is the cleaner way.

src/libos/src/fs/syscalls.rs
qzheng527: I saw a conflict about this.
In Linux manual, the count is defined as size_t which is unsigned type.
Also in th man page, there is a description "set errno EINVAL  if count is negative."

So I checked it in the Linux kernel, there is no such negative check.
https://elixir.bootlin.com/linux/latest/source/fs/read_write.c#L1175

liqinggd: Linux dose check the value:
```
int rw_verify_area(int read_write, struct file *file, const loff_t *ppos, size_t count)
{
	if (unlikely((ssize_t) count < 0))
		return -EINVAL;

```

qzheng527: @liqinggd Good catch! I will add the check back.

src/libos/crates/async-io/src/event/poller.rs
```diff
@@ -246,8 +246,8 @@ impl Drop for Poller {
         for weak_pollee in pollees.drain(..) {
             if let Some(pollee) = weak_pollee.upgrade() {
                 let mut pollers = pollee.pollers.lock();
-                let res = pollers.remove(&self_observer);
-                //assert!(res.is_some());
+                let _res = pollers.remove(&self_observer);
+                //assert!(_res.is_some());
```
guzongmin, line 250: Should we open this check in the code? Seems it could help us find bugs.

src/libos/crates/async-io/src/file/file.rs
```diff
@@ -69,6 +65,7 @@ pub trait File: Debug + Sync + Send {
         StatusFlags::empty()
     }
 
+    #[allow(unused_variables)]
```
guzongmin, line 68: Should we use the same fix as others?
fn set_status_flags(&self, _new_status: StatusFlags) -> Result<()>

src/libos/crates/async-io/src/socket/addr/c_sock_addr.rs
```diff
@@ -42,7 +42,6 @@ impl CSockAddr for libc::sockaddr_in {
 
     fn c_addr(&self) -> &[u8] {
         // Note that libc::sockaddr_in contains a useless sin_zero field of 8 bytes. Do not use it.
```
guzongmin, line 44: remove the comments, since the code it not there

src/libos/crates/async-rt/src/lib.rs
```diff
@@ -15,10 +15,9 @@
 #![feature(thread_local)]
 #![feature(const_fn_fn_ptr_basics)]
 #![feature(const_fn_trait_bound)]
-#![feature(duration_zero)]
 #![feature(duration_constants)]
-#![feature(duration_saturating_ops)]
 #![feature(get_mut_unchecked)]
+#![allow(dead_code)]
```
guzongmin, line 20: What is the dead code? Leave dead code here seems not a good thing

qzheng527: Because there are two schedulers in this crate but only one will be used.
Without this, many "unused" warnings will be output.

src/libos/crates/async-io/src/file/file.rs
qzheng527: Yes, will update

guzongmin (approved):
LGTM

PR #161, opened 2021-12-23
[libos] Fix potential file backed memory error due to mprotect

jessehui:


jessehui:
No error. This fix is useless.

PR #164, opened 2021-12-24
[libos] Fix assert issue in poller drop

qzheng527:


src/libos/crates/async-io/src/event/poller.rs
```diff
@@ -242,7 +246,7 @@ impl Drop for Poller {
             return;
         }
 
-        let self_observer = self.inner.clone() as KeyableArc<dyn Observer>;
+        let self_observer = self.observer();
         for weak_pollee in pollees.drain(..) {
             if let Some(pollee) = weak_pollee.upgrade() {
                 let mut pollers = pollee.pollers.lock();
```
guzongmin, line 252: rename the _res to res, since it is used in the next line.

guzongmin (approved):
LGTM

PR #167, opened 2021-12-27
Only send interrupt signal when task stuck in userspace over 250ms

qzheng527:
It can elimilate most of the interrupt signal sent.
The sofaboot (JDK 8/11) demo, unit test "exit_group, pthread“ can pass with this change.

### DIFF
--- a/src/pal/src/pal_interrupt_thread.c
+++ b/src/pal/src/pal_interrupt_thread.c
@@ -21,10 +21,17 @@ extern pthread_t *pal_vcpu_threads;
 extern struct occlum_pal_vcpu_data *pal_vcpu_data;
 
 static void *thread_func(void *_data) {
+    unsigned int *switch_cnts = calloc(pal_num_vcpus, sizeof(unsigned int));
+    if (switch_cnts == NULL) {
+        errno = ENOMEM;
+        return NULL;
+    }
+
     while (1) {
         struct timespec timeout = { .tv_sec = 0, .tv_nsec = 250 * MS };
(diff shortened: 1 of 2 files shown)


src/pal/src/pal_interrupt_thread.c
```diff
@@ -21,6 +21,12 @@ extern pthread_t *pal_vcpu_threads;
 extern struct occlum_pal_vcpu_data *pal_vcpu_data;
 
 static void *thread_func(void *_data) {
+    unsigned int *switch_cnts = calloc(pal_num_vcpus, sizeof(unsigned int));
```
guzongmin, line 24: Please free the buffer before function return

src/pal/src/pal_interrupt_thread.c
```diff
@@ -32,8 +38,13 @@ static void *thread_func(void *_data) {
             pthread_t vcpu_thread = pal_vcpu_threads[vcpu_i];
             struct occlum_pal_vcpu_data pal_data = pal_vcpu_data[vcpu_i];
 
-            if ( pal_data.user_space_mark == 1) {
-                pthread_kill(vcpu_thread, INTERRUPT_SIGNAL);
+            if ( pal_data.user_space_mark != 0 ) {
+                /*PAL_DEBUG("vcpu %d: previous: %d, current %d\n",
```
guzongmin, line 42: Please explain the logic in the code, show the reader how it works.

guzongmin (approved):
LGTM

PR #179, opened 2022-01-06
Fix panic happened in LTP test

qzheng527:


src/libos/src/poll/do_epoll/epoll_entry.rs
```diff
@@ -54,8 +54,8 @@ impl EpollEntry {
     }
 
     /// Get the epoll file associated with this epoll entry.
-    pub fn epoll_file(&self) -> Arc<EpollFile> {
-        self.weak_epoll.upgrade().unwrap()
+    pub fn epoll_file(&self) -> Option<Arc<EpollFile>> {
+        self.weak_epoll.upgrade()
```
guzongmin, line 58: I think the epoll entry should be removed if the epoll file is not exist.

src/libos/src/poll/do_epoll/epoll_file.rs
```diff
@@ -290,6 +290,15 @@ impl EpollFile {
     }
 }
 
+impl Drop for EpollFile {
+    fn drop(&mut self) {
+        let mut interest = self.interest.lock().unwrap();
+        for (_fd, entry) in interest.drain() {
+            entry.set_deleted();
```
guzongmin, line 297: We may need unregister_observer after set_deleted

        let file = match entry.file() {
            Some(file) => file,
            // TODO: should we warn about it?
            None => return Ok(()),
        };
        file.unregister_observer(&(entry as _)).unwrap();

src/libos/src/poll/do_epoll/epoll_file.rs
```diff
@@ -290,6 +291,20 @@ impl EpollFile {
...
+    fn drop(&mut self) {
+        let mut interest = self.interest.lock().unwrap();
+        for (fd, entry) in interest.drain() {
+            entry.set_deleted();
+            if let Some(file) = entry.file() {
+                file.unregister_observer(&(entry as _));
+            }
+
+            self.unregister_from_file(fd);
+        }
```
guzongmin, line 304:
```suggestion
    fn drop(&mut self) {
        let mut interest = self.interest.lock().unwrap();
        let fds: Vec<_> = interest
            .drain()
            .map(|(fd, entry)| {
                entry.set_deleted();
                if let Some(file) = entry.file() {
                    file.unregister_observer(&(entry as _));
                }
                fd
            })
            .collect();
        drop(interest);

        fds.iter().for_each(|fd| self.unregister_from_file(*fd));
    }
```

guzongmin: Reference EpollCtl::Del implemetation, we should hold the  interest lock and file lock separately.

src/libos/src/process/do_vfork.rs
```diff
@@ -87,6 +87,10 @@ fn restore_parent_process(mut context: *mut CpuContext, current_ref: &ThreadRef)
     let current_thread = current!();
     let current_pid = current_ref.process().pid();
 
+    // Todo: close all child created files, such as below.
+    // But parent opened files need to be filterd.
+    // current_thread.close_all_files();
```
guzongmin, line 92: Please mention the "Drop for EpollFile" to help the one to fix it when vfork implementation updated.

guzongmin (approved):
LGTM

PR #188, opened 2022-01-13
Fix overflow panic found in LTP test

qzheng527:


src/libos/src/net/syscalls.rs
```diff
@@ -416,6 +416,10 @@ pub async fn do_setsockopt(
         fd, level, optname, optval, optlen
     );
 
+    if optlen == 0 {
```
guzongmin, line 419: I think the optlen should be 0 when optval is NULL.

src/libos/src/vm/vm_manager.rs
```diff
@@ -74,14 +74,20 @@ impl VMManager {
         match addr {
             VMMapAddr::Any => {}
             VMMapAddr::Hint(addr) => {
-                let target_range = unsafe { VMRange::from_unchecked(addr, addr + size) };
+                let end = addr
+                    .checked_add(size)
+                    .ok_or_else(|| errno!(ENOMEM, "end address overflow"))?;
```
jessehui, line 79: Maybe the errno should be EOVERFLOW ?

src/libos/src/vm/vm_manager.rs
```diff
@@ -74,14 +74,20 @@ impl VMManager {
...
                 let ret = self.mmap_with_addr(target_range, options);
                 if ret.is_ok() {
                     return ret;
                 }
             }
             VMMapAddr::Need(addr) | VMMapAddr::Force(addr) => {
-                let target_range = unsafe { VMRange::from_unchecked(addr, addr + size) };
+                let end = addr
+                    .checked_add(size)
+                    .ok_or_else(|| errno!(ENOMEM, "end address overflow"))?;
```
jessehui, line 89: ditto

src/libos/src/vm/vm_range.rs
```diff
@@ -18,7 +18,10 @@ impl VMRange {
     }
 
     pub fn new_with_size(start: usize, size: usize) -> Result<VMRange> {
-        Self::new(start, start + size)
+        let end = start
```
jessehui, line 21: The above usage of VMRange::from_unchecked can be replaced with this function.

src/libos/src/net/syscalls.rs
qzheng527: Actually in current impelmentation, the optval NULL  will return EFAULT in `is_inside_user_space check`.

src/libos/src/vm/vm_manager.rs
qzheng527: From man page, EOVERFLOW implies it is happened on 32-bit architecture.
```
On 32-bit architecture together with the large file extension (i.e., using 64-bit off_t): the number of pages used for length plus number of pages used for offset would overflow unsigned long (32 bits).
```
ENOMEM is also the LTP test cases expected error number.
```
The process's maximum number of mappings would have been exceeded.
```

src/libos/src/net/syscalls.rs
guzongmin: What I think is, any code should try to explain a clear logic by itself. In this case, "If the optval is NULL, then the error code would be set to EFAULT in some place. Maybe it is correct (I am not sure). But it is not a clear truth for every reader."

guzongmin (approved):
LGTM

PR #193, opened 2022-01-19
Add untrusted allocator crate to improve untrusted memory management

jessehui:
This version only supports setting memory capacity when initialization. Support for unlimited capacity is on the way.


src/libos/crates/untrusted_allocator/src/allocator/mod.rs
```diff
@@ -0,0 +1,100 @@
...
+    }
+}
+
+impl Drop for Allocator {
+    fn drop(&mut self) {
+        debug!("[untrusted alloc] Drop allocator");
+        assert!(self.inner.lock().check_empty() == true);
+        assert!(self.inner.lock().is_free_range(&self.range));
+        let ret = unsafe { munmap(self.range.start as *mut c_void, self.range.size()) };
+        debug_assert!(ret == 0);
```
tatetian, line 98: Do you have a clear rules in mind that instruct you when to use `debug_assert` and when to use `assert`? The codebase is mixed with `debug_assert` and `assert` (for example, this function).

I recommend the following practice.
* For most defensive programming, which are intended to capture logic errors, use `debug_assert`. This way, we can add as many assertions as we want without harming the performance of release builds.
* For defense against Iago attack, always use `assert`.
* For other cases, make judgement on a case-by-case basis.

When following this guidance, most assertion statements would use `debug_assert`. By the way, this is the practice followed in std.

tatetian (commented):
The code looks promising. I think we are close to merge it.

I may not have written clearly in the review doc sent to you. I don't think we need a second crate of `untrusted_allocator`. The reason is that this second crate looks quite confusing with the `sgx-untrusted-alloc`. The two names of the crates basically mean the same. This is quite confusing when one browse the `crates` directory. And the only user of `untrusted_allocator` is `sgx-untrusted-alloc`. So having a separate crate does not facilitate any code reuse. For all these reasons, I think the `untrusted_allocator` crate should be discarded and its code should be merged into `sgx-untrusted-alloc`.

src/libos/crates/untrusted_allocator/src/allocator/mod.rs
jessehui: Got it. Thank you for summerize the rules. Done.

jessehui (commented):
Thank you for your review.

src/libos/crates/sgx-untrusted-alloc/src/untrusted_allocator/mod.rs
```diff
@@ -0,0 +1,101 @@
...
+    range: VMRange,
+    inner: Mutex<VMManager>,
+}
+
+impl Allocator {
+    // Initiate a untrusted memory allocator with size specified by user.
+    // This function is marked "unsafe" because this allocator will allocate
+    // untrusted memory which could lead to much unsafe behavior.
+    // The user must use untrusted memory with more caution.
+    pub unsafe fn new(size: usize) -> Self {
```
tatetian, line 41: I got you intention. You meant well. But still, this method should not be marked as unsafe.

Allow me to borrow an explanation from StackOverflow.

> Mark a function as `unsafe` iff the function's safety depends on its parameters or on global state. If the function is safe regardless of arguments and global state, don't mark it as `unsafe`. Whether you consider a function that uses `unsafe` internally safe is the same as whether you consider a C program safe.

You see. What `unsafe` says is: calling this function is only safe on some conditions.

In our situation, calling the `new` method is safe and secure _unconditionally_, which just creates the allocator itself. You can create an infinite number of allocators without causing any safety or security problem. So the `new` method is ok. It is safe.

What may be unsafe or insecure is using the `alloc` method, only through which the user can use untrusted memory. That is the direct cause of any potential safety or security problem. So that is the method that should  be marked unsafe.

jessehui: Got it. Thanks for your review. Done.

tatetian (approved):

PR #204, opened 2022-01-24
[tools] Make occlum package support simulation mode

qzheng527:
Merge two commits from Occlum.
f73d41 Make it accordance for occlum package name and extracted dir name
05d4c7 Add similation and debug mode occlum package check and support


volcano0dr (approved):
LGTM

PR #214, opened 2022-02-08
Update the signal stack implementation

guzongmin:
The current signal stack implementation is not align with the glibc, update the size, gap size and validate it.

### DIFF
--- a/src/libos/src/signal/do_sigreturn.rs
+++ b/src/libos/src/signal/do_sigreturn.rs
@@ -17,10 +17,7 @@ use crate::vm::{VMRange, PAGE_SIZE};
 pub fn do_rt_sigreturn() -> Result<()> {
     debug!("do_rt_sigreturn");
     let last_ucontext = {
-        let last_ucontext = PRE_UCONTEXTS.with(|ref_cell| {
-            let mut stack = ref_cell.borrow_mut();
-            stack.pop()
-        });
+        let last_ucontext = SIGNAL_CONTEXT.with(|signal_context| signal_context.borrow_mut().pop());
 
         // Handle a (very unlikely) error condition
         if last_ucontext.is_none() {
@@ -119,14 +116,6 @@ pub fn force_signal(signal: Box<dyn Signal>) {
 }
 
 fn handle_signal(signal: Box<dyn Signal>, thread: &ThreadRef) -> bool {
-    let is_sig_stack_full = PRE_UCONTEXTS.with(|ref_cell| {
-        let stack = ref_cell.borrow();
-        stack.full()
-    });
-    if is_sig_stack_full {
-        panic!("the nested signal is too deep to handle");
-    }
-
     let action = thread
         .process()
         .sig_dispositions()
@@ -314,10 +303,7 @@ fn handle_signals_by_user(
         fp_regs.clear();
     }
 
-    PRE_UCONTEXTS.with(|ref_cell| {
-        let mut stack = ref_cell.borrow_mut();
-        stack.push(ucontext).unwrap();
-    });
+    SIGNAL_CONTEXT.with(|signal_context| unsafe { signal_context.borrow_mut().push(ucontext) });
     Ok(())
 }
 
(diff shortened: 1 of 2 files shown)


src/libos/src/signal/do_sigreturn.rs
```diff
@@ -224,8 +225,21 @@ fn handle_signals_by_user(
             const BIG_ENOUGH_SIZE: u64 = 4096;
             BIG_ENOUGH_SIZE as usize
         };
-        // TODO: validate the memory range of the stack
-        unsafe { Stack::new(stack_top, stack_size)? }
+
+        let stack = unsafe { Stack::new(stack_top, stack_size)? };
```
tatetian, line 227: The code works, but should be adjusted for better readability: the range of the stack should be validated first, then create a `Stack`. 

```rust
unsafe { Stack::new(stack_top, stack_size)? };
```

This is because the `unsafe` code block above essentially says: I am totally aware that creating a new stack object from the memory range `stack_top - stack_size` to `stack_top` is risky, but I do it anyway because I am certain that the range is safe to use. Thus, validating the range after creating the stack object is illogical.

src/libos/src/signal/do_sigreturn.rs
```diff
@@ -309,6 +323,15 @@ impl Stack {
...
     }
 
+    /// Get the minimum memory range which is the superset of the stack
+    pub fn range(&self) -> VMRange {
+        VMRange::new(
+            align_down(self.bottom, PAGE_SIZE),
+            align_up(self.pointer, PAGE_SIZE),
+        )
+        .unwrap()
+    }
```
tatetian, line 328: In accordance with the comments above, the `Stack` should not have such a method.

src/libos/src/signal/do_sigreturn.rs
```diff
@@ -216,14 +217,22 @@ fn handle_signals_by_user(
                     }
                 }
             }
-            const BIG_ENOUGH_GAP: u64 = 1024;
-            let stack_top = (curr_user_ctxt.gp_regs.rsp - BIG_ENOUGH_GAP) as usize;
+            const STACK_ALIGNMENT: usize = 16;
+            const RED_ZONE: u64 = 128;
+            let stack_top = align_down(
+                (curr_user_ctxt.gp_regs.rsp - RED_ZONE).try_into().unwrap(),
+                STACK_ALIGNMENT,
```
tatetian, line 213: Using `RED_ZONE` is a nice touch, but the alignment requirement is not necessary and kind of misleading, as it gives a false sense of safety.

>        // The x86 calling convention requires rsp to be 16-byte aligned.
>        // The following allocation on stack is right before we "call" the
>        // signal handler. So we need to make sure the allocation is at least
>        // 16-byte aligned.

As stated in the comments later, the 16-byte alignment is required by the calling convention, not arbitrary x86 instructions. We only need to ensure rsp is 16-byte aligned at the entry of a function. Having a 16-byte aligned `stack_stop` at this point is useless and gives a false sense of safety to readers, since we are going to push some more data structures on the stack before making a function call.

Even if the 16-byte alignment is desired, why not enforce the property in `Stack::new`? Enforcing the alignment there is much more reliable and we can make it an explicit property of a stack.

src/libos/src/signal/do_sigreturn.rs
```diff
@@ -393,47 +379,47 @@ impl Stack {
...
+    pub fn push(&mut self, cpu_context: *mut ucontext_t) {
+        if cpu_context.is_null() {
+            panic!("the signal context is null");
+        }
 
-    pub fn push(&mut self, cpu_context: *mut ucontext_t) -> Result<()> {
-        if self.full() {
-            return_errno!(ENOMEM, "cpu context stack is full");
+        unsafe {
+            (*cpu_context).uc_link = self.context;
```
tatetian, line 403: This is cool. I wondered how unlimited nested signals are handled in Linux. But I found `ucountext_t` is poorly documented. I had no idea what is the exact purpose of `uc_link` (I though `getcontext` and `setcontext` may use it) and whether it is safe to us to use it. Could you add a reference link to explain why it is ok to use `uc_link` here?

src/libos/src/signal/do_sigreturn.rs
```diff
@@ -393,47 +379,47 @@ impl Stack {
...
-        self.count == self.stack.len()
+        SignalContext {
+            context: ptr::null_mut(),
+        }
     }
 
-    pub fn empty(&self) -> bool {
-        self.count == 0
-    }
+    pub fn push(&mut self, cpu_context: *mut ucontext_t) {
```
tatetian, line 397: This method should be marked `unsafe`. This is a common problem in my review experience. Many developers do not have a clear understanding when `unsafe` should be used.

Put simply, mark a function `unsafe` if the function requires some conditions that cannot be checked by the Rust compiler to ensure memory safety. Apparently, this function falls in the category as it has an implicit requirement that the `cpu_context` pointer is valid, pointing to valid memory region and having a proper alignment. Otherwise, the Rust code may have undefined behaviors.

tatetian (approved):
LGTM. Thanks for the contribution.

PR #215, opened 2022-02-11
[libos] Fix execve hang when the clone happens too late

jessehui:


src/libos/src/process/do_clone.rs
```diff
@@ -74,6 +74,13 @@ pub async fn do_clone(
...
     trace!("new thread sigmask = {:?}", new_thread_ref.sig_mask());
     let new_tid = new_thread_ref.tid();
+    let process = new_thread_ref.process();
+    // If the current thread is forced to exit, there is no need to let the new thread to execute.
+    if process.is_forced_to_exit() {
+        new_thread_ref.exit(process.term_status().unwrap());
+        return Ok(0);
+    }
+
     table::add_thread(new_thread_ref.clone());
```
guzongmin, line 84: When we add the new_thread into the table, we are not able to know the process status even if we already checked it. Because there is no sync mechanism.

jessehui: This check can make the "forced-exited" thread exit quickly. Once the clone function is called, the new thread will always be added to the process's thread list. The loop in `wait_for_other_threads_to_exit ` can guarantee that all other threads are exited.

guzongmin (approved):
LGTM

PR #217, opened 2022-02-15
[libos] Add stat support for disk file

liqinggd:


src/libos/src/fs/disk_file.rs
```diff
@@ -131,6 +134,25 @@ impl DiskFile {
...
+        Metadata {
+            dev: 0,
+            inode: 0x1000,
+            size: self.disk.total_bytes(),
+            blk_size: BLOCK_SIZE,
+            blocks: self.disk.total_blocks(),
+            atime: Timespec { sec: 0, nsec: 0 },
+            mtime: Timespec { sec: 0, nsec: 0 },
+            ctime: Timespec { sec: 0, nsec: 0 },
+            type_: FileType::File,
```
tatetian, line 148: Should be `FileType::BlockDevice`. User programs (e.g., fio) may use `S_ISBLK` macro to test if a value of `struct stat` type refers to a block device or not.

src/libos/src/fs/disk_file.rs
```diff
@@ -131,6 +134,25 @@ impl DiskFile {
...
         return_errno!(EINVAL, "this file does not support ioctl");
     }
+
+    pub fn metadata(&self) -> Metadata {
+        Metadata {
+            dev: 0,
+            inode: 0x1000,
+            size: self.disk.total_bytes(),
+            blk_size: BLOCK_SIZE,
+            blocks: self.disk.total_blocks(),
```
tatetian, line 144: According to the man page, the block number counts 512B blocks (i.e., sectors), not FS blocks.

```
blkcnt_t  st_blocks;  /* number of 512B blocks allocated */
```

src/libos/src/fs/disk_file.rs
```diff
@@ -131,6 +134,25 @@ impl DiskFile {
     pub fn ioctl(&self, _cmd: &mut dyn IoctlCmd) -> Result<()> {
         return_errno!(EINVAL, "this file does not support ioctl");
     }
+
+    pub fn metadata(&self) -> Metadata {
+        Metadata {
+            dev: 0,
+            inode: 0x1000,
```
tatetian, line 141: I think the value zero or a very large number would make a safer choice. The value `0x1000` could coincide with a valid inode number.

tatetian:
By the way, why CI failed? Because of this commit or not?

src/libos/src/fs/disk_file.rs
liqinggd: Many apps/libs use zero as an invalid INode number, so the id should not be zero.

src/libos/src/fs/disk_file.rs
liqinggd: Good suggestion, I presume that the bocks should be the number of block_size, but it should be the number of 512B

src/libos/src/fs/disk_file.rs
liqinggd: Actually, I have tried to use `BlockDevice`, but FIO will access it as a block device with some `ioctls`. As we already treat the block device as a file to read and write, it is reasonable to use `File`.

tatetian (approved):

PR #236, opened 2022-03-23
Add support for UTIME

Bonjourz:
1. Add syscall support for UTIME, relevant syscalls:
    a. utime
    b. utimes
    c. futimesat
    d. utimensat

2. Add test case for UTIME, other test cases are enabled in https://github.com/occlum/gvisor/pull/11

3. Update the `OcclumTimeProvider` for SEFS, which supports providing timestamp with resolution of 1 nanosecond, see `src/libos/src/time/mod.rs` for details


guzongmin (approved):
LGTM

PR #244, opened 2022-04-08
Add support for asynchronous Mutex and RwLock

jessehui:


src/libos/crates/async-rt/src/sync/mutex/mod.rs
```diff
@@ -0,0 +1,122 @@
...
+use inner::MutexInner;
+use std::boxed::Box;
+
+/// An asynchronous mutex type.
+///
+/// This is simillar to `std::sync::Mutex` but with the following differences:
+/// - `lock` method is asynchronous and will not block
+/// - the `MutexGuard` can be held across `await` calls
+pub struct Mutex<T: ?Sized> {
+    inner: Box<MutexInner>,
```
tatetian, line 14: Why Box it?

src/libos/crates/async-rt/src/sync/mutex/mod.rs
```diff
@@ -0,0 +1,122 @@
...
+            let Mutex {
+                ref inner,
+                ref data,
+            } = self;
+            unsafe { (ptr::read(inner), ptr::read(data)) }
+        };
+        mem::forget(self);
+        drop(inner);
+
+        data.into_inner()
```
tatetian, line 68: Can we do it in this simpler way?

```rust
let Mutex { _inner, data } = self;
data.into_inner()
```

src/libos/crates/async-rt/src/sync/mutex/mod.rs
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    /// Get a mutable reference to the inner data
+    pub fn get_mut(&mut self) -> &mut T {
+        unsafe { &mut *self.data.get() }
```
tatetian, line 73: Why not use `UnsafeCell::get_mut` and remove the `unsafe`?

src/libos/crates/async-rt/src/sync/mutex/mod.rs
```diff
@@ -0,0 +1,122 @@
...
+    }
+}
+
+impl<T: core::fmt::Debug> fmt::Debug for Mutex<T> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("Mutex lock")
+            .field("inner", &self.inner)
+            .field("data", unsafe { &(*self.data.get()) })
```
tatetian, line 81: This use of unsafe is not safe. Lock the data before accessing it.  See how `Debug` is implemented for `Mutex` in std. In addition, `inner` is implementation detail, I don't think it should be outputed.

src/libos/crates/async-rt/src/sync/mutex/mod.rs
```diff
@@ -0,0 +1,122 @@
...
+}
+
+impl<T: ?Sized + Default> Default for Mutex<T> {
+    fn default() -> Mutex<T> {
+        Mutex::new(Default::default())
+    }
+}
+
+impl<'a, T: ?Sized> MutexGuard<'a, T> {
+    pub fn new(lock: &'a Mutex<T>) -> MutexGuard<'a, T> {
```
tatetian, line 93: This method must be private. The users of `Mutex` must not be able to instantiate `MutexGuard` without calling `Mutex::lock` first.

src/libos/crates/async-rt/src/sync/mutex/inner.rs
```diff
@@ -0,0 +1,221 @@
...
+    }
+
+    fn get_current_status_and_set_free(&self) -> LockStatus {
+        let status = self.0.swap(LockStatus::Free as i32, Ordering::AcqRel);
+        LockStatus::try_from(status).unwrap()
+    }
+}
+
+impl LockStatus {
+    fn init() -> Self {
```
tatetian, line 201: Should be `new`

src/libos/crates/async-rt/src/sync/rwlock/mod.rs
```diff
@@ -0,0 +1,150 @@
+use super::*;
+
+mod inner;
+
+use inner::RwLockInner;
+
+/// A readers-writer lock implementation which allows many readers or at most one writer at the same time.
+pub struct RwLock<T: ?Sized> {
+    inner: Box<RwLockInner>,
```
tatetian, line 9: Again, why box?

src/libos/crates/async-rt/src/sync/mutex/inner.rs
```diff
@@ -0,0 +1,221 @@
...
+                    Errno::EINTR => continue,
+                    _ => {
+                        return_errno!(error.errno(), "futex wait error")
+                    }
+                }
+            }
+        }
+    }
+
+    pub(super) fn unlock(&self) -> Result<()> {
```
tatetian, line 119: When unlock could fail?

src/libos/crates/async-rt/src/sync/rwlock/inner.rs
```diff
@@ -0,0 +1,476 @@
+// Copyright (C) 2020 Ant Financial Services Group. All rights reserved.
+
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions
+// are met:
+
```
tatetian, line 6: I think the whole copyright block should not be broken by empty line. Break it by empty comment line instead.

src/libos/crates/async-rt/src/task/mod.rs
```diff
@@ -23,7 +23,7 @@ pub fn spawn<T: Send + 'static>(future: impl Future<Output = T> + 'static + Send
     SpawnOptions::new(future).spawn()
 }
 
-pub fn block_on<T: Send + 'static>(future: impl Future<Output = T> + 'static + Send) -> T {
+pub fn block_on<'a, T: Send + 'static>(future: impl Future<Output = T> + 'static + Send) -> T {
```
tatetian, line 26: Is `'a` used anywhere?

src/libos/crates/async-rt/src/sync/mutex/mod.rs
```diff
@@ -0,0 +1,122 @@
...
+
+    /// Get a mutable reference to the inner data
+    pub fn get_mut(&mut self) -> &mut T {
+        unsafe { &mut *self.data.get() }
+    }
+}
+
+impl<T: core::fmt::Debug> fmt::Debug for Mutex<T> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("Mutex lock")
```
tatetian, line 79: Just print the type name, which is `Mutex`.

src/libos/crates/async-rt/src/sync/mutex/inner.rs
```diff
@@ -0,0 +1,221 @@
...
+        // `get_current_status_and_set_free` will make sure this happens before. Thus, `Relaxed` can be used here.
+        let waiters = self.waiters.load(Ordering::Relaxed);
+        let previous_status = self.status.get_current_status_and_set_free();
+
+        // Also check previous status in case the waiter number overflows.
+        if waiters > 0 || previous_status.is_locked_with_waiters() {
```
tatetian, line 125: Why checking `is_locked_with_waiters` is going to help with overflow? Could you explain?

src/libos/crates/async-rt/src/sync/mutex/inner.rs
```diff
@@ -0,0 +1,221 @@
+/// This implementation makes reference to musl libc's design but only keep the basic
+/// functionality. Recursive, error-checking, priority-inheritance or robust-list
+/// are not supported yet.
+// use super::super::futex::{FutexPtr, FUTEX_BUCKETS};
```
tatetian, line 4: Remove all futex comments and fix error message containing "futex".

tatetian (commented):
The PR looks very promising. Good job on removing futexes. The lock logic seems correct to me. I like the inclusion of musl libc copyright. It is decent for us to do so. But I think the mutex implementation could be refactored to get rid of the impact from musl libc.

Many of my comments for `Mutex` also apply to `RwLock`. I don't want to repeat myself. I hope you can also make changes to `RwLock` when it is fit to do so.

I need a bit more time of diving in the usage of `Ordering`. I am not entirely sure if they are used correctly or in the most relaxed way. Maybe we could talk about it offline.

Last, given the importance of the lock primitives, it is important to include unit tests. Even benchmarks, if you ask. But I think it is ok to add benchmarks in future PRs. The bottom line is that there should be some unit tests.

Thanks again. This is good work.

src/libos/crates/async-rt/src/lib.rs
```diff
@@ -199,6 +206,195 @@ mod tests {
         .spawn()
     }
 
+    #[test]
+    fn test_mutex() {
```
tatetian, line 210: Unit tests for mutexes should be moved to mutex module. So are rwlock tests. We cannot write all unit tests in `lib.rs`.  Most unit tests in `lib.rs` are there only for historical reasons; they should be put under appropriate modules, instead of the root modules.

src/libos/crates/async-rt/src/sync/mutex/inner.rs
```diff
@@ -0,0 +1,221 @@
...
+use super::*;
```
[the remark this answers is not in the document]
jessehui: I understand your point. But I want to add another legitimate usage that when the import types are commonly used std/core types like `Box`, `Atomic` etc, these can be put to `::*`. For other types, whether defined somewhere in the current crate or from external crates, should not be put into `::*`. I think this is acceptable. Modification done.

src/libos/crates/async-rt/src/sync/mutex/inner.rs
```diff
@@ -0,0 +1,221 @@
...
+#[derive(Debug)]
+struct AtomicLockStatus(AtomicI32);
+
+// The status code keeps the same as the implementation in musl.
+#[derive(Debug, Copy, Clone)]
+#[repr(i32)]
+enum LockStatus {
+    Free = 0,
+    Locked = Errno::EBUSY as i32,
+    LockedWithWaiters = -2147483632, // 0x8000_0010 = (EBUSY | 0x8000_0000)
```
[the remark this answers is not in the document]
jessehui: Re-assign values. Done.

src/libos/crates/async-rt/src/sync/mutex/inner.rs
```diff
@@ -0,0 +1,221 @@
...
+        let locked_waiter_queue = self.waiter_queue.lock();
+        // Check the status value again
+        if !self.status.is_locked_with_waiters() {
+            return_errno!(EAGAIN, "the status has changed");
+        }
+
+        locked_waiter_queue.enqueue(&mut waiter);
+
+        drop(locked_waiter_queue);
+        waiter.wait().await
```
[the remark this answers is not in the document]
jessehui: Add `dequeue` after `wait`.

src/libos/crates/async-rt/src/sync/mutex/mod.rs
jessehui: Unnecessary. Removed.

src/libos/crates/async-rt/src/sync/mutex/mod.rs
jessehui: Done.

src/libos/crates/async-rt/src/sync/mutex/mod.rs
jessehui: Done.

src/libos/crates/async-rt/src/task/mod.rs
jessehui: Removed.

src/libos/crates/async-rt/src/sync/rwlock/inner.rs
jessehui: Done.

src/libos/crates/async-rt/src/sync/mutex/inner.rs
```diff
@@ -0,0 +1,221 @@
...
+            }
+
+            // In try_set_lock_with_waiters, `AcqRel` will make sure this increment happens before. Thus, `Relaxed` can be used here.
+            self.waiters.fetch_add(1, Ordering::Relaxed);
+
+            // Ignore the result here. If the state transition fails, the next wait will not block.
+            self.status
+                .try_set_lock_with_waiters()
+                .map_err(|e| errno!(e.errno(), "failed to set lock status"))
+                .ok();
```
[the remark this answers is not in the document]
tatetian: If results are to be ignored, then the right to do it is as follows.

```rust
let _ = self.status.try_set_lock_with_waiters();
```

In addition, printing errors and warnings is not a good idea. Because failures to grab a lock are harmless, trivial, and may be very frequent in some cases. There is no point in printing such messages. They are going to be noises.

src/libos/crates/async-rt/src/sync/mutex/inner.rs
```diff
@@ -0,0 +1,221 @@
...
+
+        // Also check previous status in case the waiter number overflows.
+        if waiters > 0 || previous_status.is_locked_with_waiters() {
+            self.wake_one_waiter();
+        }
+
+        Ok(())
+    }
+
+    async fn wait(&self) -> Result<()> {
```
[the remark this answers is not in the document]
tatetian: My understanding is that there is a single caller of this method. And this caller does not care about the result of this method: it always retries regardless of the outcome. Since the caller and callee are all written by you, why bother return a useless result?

src/libos/crates/async-rt/src/sync/mutex/inner.rs
tatetian: I didn't get it. You are basically saying that having `waiters` does not really matter (because it can overflow anyway) and all we need to do is to check `is_locked_with_waiters`. If so, why bother keeping the waiters counter? We can just check `is_locked_with_waiters` alone.

src/libos/crates/async-rt/src/sync/mutex/inner.rs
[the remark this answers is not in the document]
jessehui: Done.

src/libos/crates/async-rt/src/sync/mutex/inner.rs
jessehui: The `waiters` checking is the main check. `is_locked_with_waiters` is only to protect the situation that the waiter counter overflows to 0. 
`is_locked_with_waiters` can't be used as the main check because the status will be `Locked` when thread 1 `unlock` and `wake` up thread 2 and thread 2 `try_lock` succeeds. Thus, when thread 2 `unlock`, the `is_locked_with_waiters` check will fail and other waiters will not be woken up. 
However, `is_locked_with_waiters` check can guard the situation that the waiter counter overflows to 0 because this situation only exists once. And in the next unlock, the waiter counter will not be 0 anymore.

src/libos/crates/async-rt/src/sync/mutex/inner.rs
```diff
@@ -0,0 +1,199 @@
...
+        let previous_status = self.status.get_current_status_and_set_free();
+
+        if waiters != 0 {
+            self.wake_one_waiter();
+            return;
+        }
+
+        if previous_status.is_locked_with_waiters() {
+            panic!("the waiter number overflows to 0");
+        }
```
tatetian, line 105: This is incorrect. It is possible that `waiters == 0 && previous_status.is_locked_with_waiters()` without overflow. You can figure it out yourself... in no time.

jessehui: Got it. Remove this check. Done.

tatetian (approved):

PR #245, opened 2022-04-12
Add hostname and hosts file

ClawSeven:
- Support hosts and hostname file in enclave
- Refine copied file struct
- Modify uname hard code nodename


guzongmin (approved):
LGTM

PR #250, opened 2022-05-09
[demo] Add Filebench demo

lucassong-mh:
Filebench tool can generate **real-world workloads** to test **file system** and **block device**
1. Add demo with four workloads ([Sworndisk paper evaluation](https://github.com/occlum/lets-go-sworn-disk/issues/48#666))
2. Add Filebench_test to ci:demo_test workflow

### DIFF
--- a/demos/filebench/dl_and_build_filebench.sh
+++ b/demos/filebench/dl_and_build_filebench.sh
@@ -1,0 +1,40 @@
+#!/bin/bash
+set -e
+
+BLUE='\033[1;34m'
+NC='\033[0m'
+echo -e "${BLUE}Start installing dependencies.${NC}"
+
+DEPS="bison flex"
+
+apt-get update
+apt-get install -y $DEPS
+echo -e "${BLUE}Finish installing dependencies.${NC}"
+
+echo -e "${BLUE}Start building filebench from tarball.${NC}"
+
+# Download release tarball
+VERSION="1.5-alpha3"
+TARBALL="filebench-${VERSION}.tar.gz"
+rm -f ${TARBALL}
+wget https://github.com/filebench/filebench/releases/download/${VERSION}/${TARBALL}
+rm -rf filebench && mkdir filebench
+tar -zxvf filebench-${VERSION}.tar.gz -C filebench --strip-components 1
+pushd filebench
+
+./configure
+popd
+
+# Make modification to
+# 1. Replace fork to vfork
+# 2. Prepare shared memory region for child processes
+# 3. Disable SYSV semaphores
+patch -s -p0 < filebench.patch
+
+pushd filebench
+# Build and install filebench tool
+make -j$(nproc)
+make install
+echo -e "${BLUE}Finish building filebench from tarball.${NC}"
+
+popd
--- a/demos/filebench/workloads/oltp.f
+++ b/demos/filebench/workloads/oltp.f
@@ -1,0 +1,81 @@
+#
+# CDDL HEADER START
+#
+# The contents of this file are subject to the terms of the
+# Common Development and Distribution License (the "License").
+# You may not use this file except in compliance with the License.
+#
+# You can obtain a copy of the license at usr/src/OPENSOLARIS.LICENSE
+# or http://www.opensolaris.org/os/licensing.
+# See the License for the specific language governing permissions
+# and limitations under the License.
+#
+# When distributing Covered Code, include this CDDL HEADER in each
+# file and include the License file at usr/src/OPENSOLARIS.LICENSE.
+# If applicable, add the following below this CDDL HEADER, with the
+# fields enclosed by brackets "[]" replaced with your own identifying
+# information: Portions Copyright [yyyy] [name of copyright owner]
+#
+# CDDL HEADER END
+#
+#
+# Copyright 2009 Sun Microsystems, Inc.  All rights reserved.
+# Use is subject to license terms.
+#
+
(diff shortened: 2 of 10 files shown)


demos/filebench/dl_and_build_filebench.sh
```diff
@@ -0,0 +1,42 @@
...
+# 1. Replace fork to vfork
+# 2. Prepare shared memory region for child processes
+# 3. Disable SYSV semaphores
+sed -i '93d' procflow.c
+sed -i '93 i  if ((pid = vfork()) < 0) {' procflow.c
+sed -i '418 i  filebench_shm = (filebench_shm_t *)shmaddr;' ipc.c
+sed -i '403,417d' ipc.c
+
+./configure
+sed -i '298 i  #undef HAVE_SYSV_SEM' config.h
```
tatetian, line 35: Using sed to modify the source code does not seem to be maintainable. How about having diff file? So that in this script we can simply apply the patch.

demos/filebench/workloads/oltp.f
```diff
@@ -0,0 +1,86 @@
...
+define fileset name=logfile,path=$dir,size=$logfilesize,entries=$nlogfiles,dirwidth=1024,prealloc=100,reuse
+
+define process name=lgwr,instances=1
+{
+  thread name=lgwr,memsize=$memperthread
+  {
+    flowop aiowrite name=lg-write,filesetname=logfile,
+        iosize=256k,random,directio=$directio,dsync
+    flowop aiowait name=lg-aiowait
+    flowop semblock name=lg-block,value=3200,highwater=1000
```
tatetian, line 54: Should we use aio when Occlum does not support it? I think aio is a Linux feature, not a user-space library feature. Any other features that Occlum does not support and should be worked around?

demos/filebench/workloads/oltp.f
```diff
@@ -0,0 +1,86 @@
...
+  thread name=shadow,memsize=$memperthread
+  {
+    flowop read name=shadowread,filesetname=datafiles,
+        iosize=$iosize,workingset=$workingset,random,opennext,directio=$directio
+    flowop hog name=shadowhog,value=$usermode
+    flowop sempost name=shadow-post-lg,value=1,target=lg-block,blocking
+    flowop sempost name=shadow-post-dbwr,value=1,target=dbwr-block,blocking
+    flowop eventlimit name=random-rate
+  }
+}
```
tatetian, line 82: I thought multi-process workloads cannot be supported since Occlum does not have vfork. So how can Occlum run such multi-process workloads?

tatetian (commented):
The PR Looks good. I raised some minor issues that I would like to discuss with yoou.

demos/filebench/dl_and_build_filebench.sh
lucassong-mh: Make sense to me. I had thought that "diff+patch" is a git feature.

demos/filebench/workloads/oltp.f
lucassong-mh: It seems that non-supported features all appear in "oltp", which are `aio` and `semaphore` related flowops.

demos/filebench/workloads/oltp.f
lucassong-mh: I found `vfork` implementation both in Occlum and NGO. And tests go well under multi-process configs after I replace `fork` to `vfork`.
And `vfork` is a "must-option" since even if we set `instances=1`, filebench main process still forks one child to run the test.

demos/filebench/workloads/oltp.f
lucassong-mh: Should we modify `oltp` to adapt Occlum? For example we use `randomwrite+fsync` rather than `aiowrite+aiowait`.

demos/filebench/workloads/oltp.f
lucassong-mh: Replacing to `vfork` can provide concurrency since filebench use `vfork`+`execve`.
[vfork man page](https://man7.org/linux/man-pages/man2/vfork.2.html):

> the calling thread is suspended until the child terminates (either normally, by calling _exit(2), or abnormally, after delivery of a fatal signal), or it makes a call to **execve(2)**.

tatetian: I see. I thought filebench uses fork alone, without execve. Are you sure the replacing fork with vfork is correct? I mean vfork has some limitations (a child process cannot modify the parent's variables). If so, then filebench should have a complete multi-process support running on Occlum, right? And if yes, then we don't need to rewrite .f files to avoid multi-process usage.

demos/filebench/workloads/oltp.f
tatetian: I am not entirely sure. You are in a better position to make a sensible decision than I do :)

demos/filebench/workloads/oltp.f
lucassong-mh: When I enable DEBUG mode, log shows multi-process executes operations concurrently and the results are normally consistent with results on Linux.
And yes, no need to rewrite this part in .f files.

demos/filebench/workloads/oltp.f
lucassong-mh: I've already replace `aiowrite` to `write` with same parameters. `oltp` works fine now.

tatetian (approved):

PR #251, opened 2022-05-10
Add connect_poller method to reduce lock when calling poll in a loop

jessehui:


src/libos/crates/async-io/src/event/poller.rs
```diff
@@ -84,6 +84,21 @@ impl Pollee {
         revents
     }
 
+    pub fn fast_poll(
```
guzongmin, line 87: Do not get the point. Why don't you use poll (mask, none) in the code directly?

tatetian: The intention of adding `fast_poll` is good. But the API does not feel good to me.

**The first problem is that the API is NOT easy-to-use.** The user has to pass in a `first_time` variable. And the user must not accidentally modify it. If we have `fast_poll` method, I hope it can be more automatic and transparent to users, hiding the decision condition (which is `first_time`) from the user.

**The second problem is that the API is prone to misuse.** After the first call, `fast_poll` switches to `poll`, assuming that the first call must have connected the poller and pollee. But this assumption is not true: _the `poll` method connects a poller and a pollee only when the returned events are empty_. In most use cases, if the user code finds the return events are non-empty, then it will stop polling; so `fast_poll` is ok. But if, for some reasons, the user code decides to poll again even when the non-empty events are returned in the last time, then using `fast_poll` may lead to wait forever.

jessehui: @guzongmin @tatetian Thank you for your review. I try a different approach to separate the `connect` and `poll` semantics for users to call based on their needs.

volcano0dr (approved):

tatetian:
This PR changes the contract that is previously provided by `Poller::poll` without updating the document. 

```rust
    /// This operation is _atomic_ in the sense that either some interesting
    /// events are returned or the poller is registered (if a poller is provided).
```

The atomicity guarantee is now broken due to the changes. The new implementation always "connect" the poller and the pollee regardless whether some events can be returned. So the doc should be updated.

A even more important question is whether the atomicity should be given up.  Atomicity is desirable because if `pollee.poll(mask, Some(&poller))` can return some events, then the `poller` won't get connected to the pollee, which means avoiding all the costs associated with it.

PR #261, opened 2022-06-06
Identify and fix performance issues in async code

tatetian:
This PR is intended to identify and fix performance issues regarding the async runtime, async I/O, io_uring, and block devices.

As the first steps, this PR adds a series of benchmarks. Here are some of the results.
* Benchmark results of `async_rt` primitives
image
* Benchmark results of `Channel<u8>` in `async_io`
image
* Benchmark results of several types of disks in `sgx-disk`
image

Based on the observations acquired from the benchmarks, this PR fixes several performance issues in async_rt and io_uring_callack.
* [[async-rt] Fix the premature sleep of the executor](https://github.com/tatetian/ngo/commit/6e28598693cd9d84f493d41492cde3a00323634b)
* [[io-uring-callback] Improve the efficiency of completion polling](https://github.com/tatetian/ngo/commit/7297664b245ed7c7965e1f74777d1c6b92a40a67)
* [[io-uring-callback] Enable kernel polling mode](https://github.com/tatetian/ngo/commit/cb4b5dcf1a987d4ed28aa6df6081de0b70eee2a6)

### DIFF
--- a/src/libos/crates/Cargo.toml
+++ b/src/libos/crates/Cargo.toml
@@ -13,18 +13,22 @@ members = [
     "object-id",
     "sgx-disk",
     "sgx-untrusted-alloc",
-    "vdso-time",
+    "vdso-time"
 ]
 
 # Default members can run on Linux; non-default members can only run inside SGX.
 default-members = [
     "async-rt",
+    "async-io",
+    "block-device",
+    "host-socket",
     "inherit-methods-macro",
     "io-uring-callback",
     "keyable-arc",
     "new-self-ref-arc",
-    "vdso-time",
-    "host-socket"
+    "object-id",
+    "sgx-disk",
+    "vdso-time"
 ]
 
 exclude = [ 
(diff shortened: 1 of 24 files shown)


tatetian:
Need to figure out why this test from gvisor failed. It seems that this test failure can be reproduced reliably. I ran the test twice; both times failed.

```
[  FAILED  ] SharedPrivate/PrivateAndSharedFutexTest.Wait_BitsetTimeout/0, where GetParam() = false (1000 ms)
```

jessehui:
> The failure test from gvisor says:
> 
> ```
> test/syscalls/linux/futex.cc:198: Failure
> Expected: (timer.Duration()) >= (kTimeout), actual: 999.874883ms vs 1s
> [  FAILED  ] SharedPrivate/PrivateAndSharedFutexTest.Wait_BitsetTimeout/0, where GetParam() = false (1000 ms)
> ```

This is a common error. The timer implementation can only provide `ms` accuracy. 
Futex test error will be ignored in the future: https://github.com/occlum/gvisor/pull/13

tatetian:
> 

Thanks for the input. I checked the last few commits; they didn't fail on this test. But I ran the test twice on this PR; both times failed. So I guess this PR increases the odds of the test failure.

volcano0dr (approved):

Issue #262, opened 2022-06-06
Improve the scalability of SyncIoDisk


tatetian:
The `SyncIoDisk` in `sgx-disk` is not scalable on concurrency.

![image](https://user-images.githubusercontent.com/568208/172077666-09e9538f-7200-4709-b3dd-db1eace1f6ec.png)

The problem is that there is a lock hold while doing I/O. This lock can be removed if using `read_at` and `write_at` API.

![image](https://user-images.githubusercontent.com/568208/172077801-a0c09565-5595-4505-b643-39557bd7acf2.png)

lucassong-mh:
Workaround: Use `std::os::unix::fs::FileExt::read_at` and `write_at` API, remove the lock.
Bench result:
![image](https://user-images.githubusercontent.com/22837133/172284685-ce782a27-513a-4e6a-a0cd-90b983a3bfec.png)
Code:
![image](https://user-images.githubusercontent.com/22837133/172285541-51972777-478c-4a47-8699-f6af584b683f.png)

tatetian:
This is good perf improvement. But the implementation contains a bug... The offset should be calculated as follows.

```
offset += buf.as_slice().len()
```

Another thing is that the implementation should better use `pwritev` so that the number of OCalls is always one per request, instead of one per buffer.

lucassong-mh:
Improvement: Use `libc::ocall::preadv64`and`pwritev64` API.
Bench result:
![image](https://user-images.githubusercontent.com/22837133/172518668-d4933089-bd62-41dd-81ff-fe93c6099f33.png)
Code:
![image](https://user-images.githubusercontent.com/22837133/172518801-2d0aa8f7-ad09-4eb2-9ff5-a547eb187663.png)

tatetian:
Great. Now we can submit a pull request.

PR #264, opened 2022-06-09
[sgx-disk] Improve SyncIoDisk scalability

lucassong-mh:
`SyncIoDisk` scalability  issue: #262 

Improvement:
- Use `libc::ocall::preadv64` and `pwritev64` API for host file reads/writes
- Remove the mutex on file

### DIFF
--- a/src/libos/crates/sgx-disk/src/host_disk/sync_io_disk.rs
+++ b/src/libos/crates/sgx-disk/src/host_disk/sync_io_disk.rs
@@ -1,7 +1,8 @@
-use block_device::{BioReq, BioSubmission, BioType, BlockDevice};
+use block_device::{BioReq, BioSubmission, BioType, BlockBuf, BlockDevice};
 use fs::File;
 use std::io::prelude::*;
 use std::io::{IoSlice, IoSliceMut, SeekFrom};
+use std::os::unix::io::AsRawFd;
 use std::path::{Path, PathBuf};
 
 use super::OpenOptions;
@@ -20,7 +21,7 @@ use crate::HostDisk;
 /// It is recommended to use `IoUringDisk` for an optimal performance.
 #[derive(Debug)]
 pub struct SyncIoDisk {
-    file: Mutex<File>,
+    file: File,
     path: PathBuf,
     total_blocks: usize,
     can_read: bool,
@@ -33,23 +34,10 @@ impl SyncIoDisk {
             return Err(errno!(EACCES, "read is not allowed"));
         }
 
+        let fd = self.file.as_raw_fd();
         let (offset, _) = self.get_range_in_bytes(&req)?;
 
-        let mut file = self.file.lock().unwrap();
-        file.seek(SeekFrom::Start(offset as u64))?;
-        let read_len = req.access_mut_bufs_with(|bufs| {
-            let mut slices: Vec<IoSliceMut<'_>> = bufs
-                .iter_mut()
-                .map(|buf| IoSliceMut::new(buf.as_slice_mut()))
-                .collect();
-
-            // TODO: fix this limitation
-            const LINUX_IOVS_MAX: usize = 1024;
-            debug_assert!(slices.len() < LINUX_IOVS_MAX);
-
-            file.read_vectored(&mut slices)
-        })?;
-        drop(file);
+        let read_len = req.access_mut_bufs_with(|bufs| self.preadv(fd, bufs, offset as i64))?;
 
         debug_assert!(read_len / BLOCK_SIZE == req.num_blocks());
         Ok(())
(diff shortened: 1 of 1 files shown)


src/libos/crates/sgx-disk/src/host_disk/sync_io_disk.rs
```diff
@@ -104,6 +89,58 @@ impl SyncIoDisk {
...
+        let read_len = unsafe {
+            #[cfg(not(feature = "sgx"))]
+            let read_len = libc::preadv64(fd, iovecs_ptr, iovecs_len, offset);
+            #[cfg(feature = "sgx")]
+            let read_len = libc::ocall::preadv64(fd, iovecs_ptr, iovecs_len, offset);
+            read_len
+        };
+        if read_len < 0 {
+            return_errno!(Errno::from(libc_errno() as u32), "libc::preadv64 error");
+        }
```
tatetian, line 114: Using the result from ocall without check is prone to Iago attacks.

After examining [the code of `libc::ocall::preadv64`](https://github.com/occlum/incubator-teaclave-sgx-sdk/blob/c2d700f8a622745204bf7d270fe85f2b9d77925e/sgx_libc/src/linux/x86_64/ocall.rs#L1845), I reach the conclusion that we do need to do the extra sanity check that `read_len` is not larger than the combined length of iovecs.

src/libos/crates/sgx-disk/src/host_disk/sync_io_disk.rs
```diff
@@ -104,6 +89,58 @@ impl SyncIoDisk {
...
+        let write_len = unsafe {
+            #[cfg(not(feature = "sgx"))]
+            let write_len = libc::pwritev64(fd, iovecs_ptr, iovecs_len, offset);
+            #[cfg(feature = "sgx")]
+            let write_len = libc::ocall::pwritev64(fd, iovecs_ptr, iovecs_len, offset);
+            write_len
+        };
+        if write_len < 0 {
+            return_errno!(Errno::from(libc_errno() as u32), "libc::pwritev64 error");
+        }
```
tatetian, line 140: Need to check `write_len`.

src/libos/crates/sgx-disk/src/host_disk/sync_io_disk.rs
```diff
@@ -33,23 +34,16 @@ impl SyncIoDisk {
...
         let read_len = req.access_mut_bufs_with(|bufs| {
-            let mut slices: Vec<IoSliceMut<'_>> = bufs
-                .iter_mut()
-                .map(|buf| IoSliceMut::new(buf.as_slice_mut()))
-                .collect();
-
             // TODO: fix this limitation
             const LINUX_IOVS_MAX: usize = 1024;
-            debug_assert!(slices.len() < LINUX_IOVS_MAX);
+            debug_assert!(bufs.len() < LINUX_IOVS_MAX);
```
tatetian, line 43: Put this check in writev and readv. Because that is where buffers are converted to iovecs and thus is where the assertion makes most sense.

tatetian (commented):
Looks great. There are only some minor issues.

src/libos/crates/sgx-disk/src/host_disk/sync_io_disk.rs
lucassong-mh: Should we return `EINVAL` if `read_len`/`write_len` is not exactly equal to the combined length of all the iovecs?
Or we send an error log and return an `Ok` (which means allow partial preads/pwrites succeed)?

src/libos/crates/sgx-disk/src/host_disk/sync_io_disk.rs
lucassong-mh: Done

src/libos/crates/sgx-disk/src/host_disk/sync_io_disk.rs
lucassong-mh: To confirm, Partial reads/writes are unlikely to happen since buffers and offset are all within legal range of the host file (file size is truncated to size of total blocks when created)

src/libos/crates/sgx-disk/src/host_disk/sync_io_disk.rs
```diff
@@ -104,6 +77,93 @@ impl SyncIoDisk {
...
+                    }
+                })
+                .collect(),
+        );
+        let (iovecs_ptr, iovecs_len) =
+            { (iovecs.as_ptr() as *const libc::iovec, iovecs.len() as i32) };
+
+        // TODO: fix this limitation
+        const LINUX_IOVS_MAX: i32 = 1024;
+        debug_assert!(iovecs_len < LINUX_IOVS_MAX);
```
tatetian, line 146: My second though is that we should remove this assert.

The man page of `pwritev` states.

> EINVAL
> The sum of the iov_len values overflows an ssize_t value. Or, the vector count iovcnt is less than zero or greater than the permitted maximum.

> Linux notes
> POSIX.1-2001 allows an implementation to place a limit on the number of items that can be passed in iov. An implementation can advertise its limit by defining IOV_MAX in <[limits.h](https://linux.die.net/include/limits.h)> or at run time via the return value from sysconf(_SC_IOV_MAX). On Linux, the limit advertised by these mechanisms is 1024, which is the true kernel limit. However, the glibc wrapper functions do some extra work if they detect that the underlying kernel system call failed because this limit was exceeded. In the case of readv() the wrapper function allocates a temporary buffer large enough for all of the items specified by iov, passes that buffer in a call to [read](https://linux.die.net/man/2/read)(2), copies data from the buffer to the locations specified by the iov_base fields of the elements of iov, and then frees the buffer. The wrapper function for writev() performs the analogous task using a temporary buffer and a call to [write](https://linux.die.net/man/2/write)(2).

Linux returns `EINVAL` if the length of iovecs is too long. Or the glibc wrapper can combines the iovecs into a big buffer. As this error can be captured and returned to the user at runtime, we should give an opportunity to handle it instead of panic.

lucassong-mh: Make sense to me. Done

tatetian (approved):

PR #279, opened 2022-07-20
[sgx-untrusted-alloc] Add support for infinite capacity of untrusted memory

jessehui:


volcano0dr:
The current implementation can cause infinite expansion of untrusted memory. I think we should consider implementing chunk recovery.

src/libos/crates/sgx-untrusted-alloc/src/untrusted_allocator/mod.rs
```diff
@@ -13,86 +13,99 @@ const PAGE_SIZE: usize = 4096;
 // The address of a block returned by malloc or realloc in GNU systems is always a multiple of eight (or sixteen on 64-bit systems).
 const DEFAULT_ALIGNMENT: usize = 16;
 
+const DEFAULT_CHUNK_SIZE: usize = 4 * 1024; // FOR TEST ONLY
```
volcano0dr, line 16: Please modify it to 16M.

jessehui: Done. Thanks for you review.

volcano0dr (approved):
LGTM

PR #307, opened 2022-10-12
Ngo boot update

qzheng527:
Main changes as below.
1. Combine two internal config json as one.
2. Use KSS ISV Family ID to indicate the measurement of app file system.
3. Clean up unnecessary internal files.
4. Add document for NGO boot flow.
5. Add sealing policy setting for users.


.github/workflows/demo_test.yml
```diff
@@ -32,15 +32,6 @@ jobs:
             SGX_MODE=SIM occlum build;
             occlum run /bin/hello_world"
 
-    - name: C with encrypted image test
```
liqinggd, line 35: Since we have deleted this case, how to test if the fs image with a key can work?  Is there another test case that can substitute this one?

src/libos/src/fs/rootfs.rs
```diff
@@ -242,3 +245,109 @@ fn open_or_create_sefs_according_to(
...
+            from_user::clone_cstring_safely(path)?
+                .to_string_lossy()
+                .into_owned(),
+        ))
+    };
+
+    Ok(path)
+}
+
+pub fn gen_config_app(config: &user_rootfs_config) -> Result<ConfigApp> {
```
liqinggd, line 281: I think this implementation is not Object-Oriented, you should implement a `From` for `user_rootfs_config` to transform into `ConfigApp`.  BTW, the structs here should be defined in `config.rs` instead of this file.

demos/runtime_boot/build_content.sh
```diff
@@ -0,0 +1,90 @@
+#!/bin/bash
+set -e
+
+script_dir="$( cd "$( dirname "${BASH_SOURCE[0]}"  )" >/dev/null 2>&1 && pwd )"
+
+export BASH_DEMO_DIR="${script_dir}/../bash"
+export INIT_DIR="${script_dir}/init"
+
+UNIONFS_DIR="${script_dir}/gen_rootfs_instance/mnt_unionfs"
+ENCRIP_KEY="c7-32-b3-ed-44-df-ec-7b-25-2d-9a-32-38-8d-58-61"
```
liqinggd, line 10: The work spelling is not correct

src/libos/src/config.rs
```diff
@@ -128,15 +126,23 @@ pub struct ConfigUntrustedUnixSock {
...
     pub options: ConfigMountOptions,
 }
 
-#[derive(Debug, PartialEq)]
+#[derive(Clone, Debug)]
+pub struct ConfigApp {
+    pub entry_points: Vec<PathBuf>,
+    pub stage: String,
+    pub mount: Vec<ConfigMount>,
+    pub encrypted: bool,
```
liqinggd, line 142: I can not see the benefit of this newly added field `encrypted`. 

The `init` program should pass `key`  to `mount_rootfs` syscall for the encrypted image, if the value is not passed or the key is wrong, it will fail.

So this filed in Occlum.json seems to be useless

liqinggd (changes requested):

.github/workflows/demo_test.yml
qzheng527: The new added runtime_boot demo can cover the runtime mount encrypted rootfs image.
The remote_attestation/init_ra_flow can cover the general building encrypted image and acquiring key in init process. (In the HW CI)

demos/runtime_boot/build_content.sh
qzheng527: Good catch, I will fix it

src/libos/src/fs/rootfs.rs
qzheng527: I will see how to address this.

src/libos/src/config.rs
qzheng527: > encrypted

You are right about the "useless" . The reason I passed "encrypted" from build process to libos was trying to give user a way (checking the generared json file) to tell why the image could be boot successfully. But since no key, no successful boot for encrypted image, then perhaps it is useless.

liqinggd: If we keep this filed, then maybe Occlum can check this value with key parameter and early return error before mounting FS.

.github/workflows/demo_test.yml
liqinggd: I think the runtime_boot demo should be added into the demo test CI

src/libos/src/config.rs
qzheng527: I will update the commit with key&type check in the entry of do_mount_rootfs for general boot case.

src/libos/src/config.rs
```diff
@@ -446,4 +493,131 @@ struct InputConfigMountOptions {
...
+pub struct user_rootfs_config {
+    upper_layer_path: *const i8,
+    lower_layer_path: *const i8,
+    entry_point: *const i8,
+    hostfs_source: *const i8,
+    hostfs_target: *const i8,
+}
+
+impl user_rootfs_config {
+    pub fn from_raw_ptr(ptr: *const user_rootfs_config) -> Result<user_rootfs_config> {
```
liqinggd, line 527: this `from_raw_ptr ` method should be unsafe, and it should not return `Result<T>`, because this method cannot fail.

src/libos/src/fs/syscalls.rs
```diff
@@ -570,20 +570,31 @@ pub async fn do_ioctl(fd: FileDesc, cmd: u32, argp: *mut u8) -> Result<isize> {
 
 pub async fn do_mount_rootfs(
     key_ptr: *const sgx_key_128bit_t,
-    occlum_json_mac_ptr: *const sgx_aes_gcm_128bit_tag_t,
+    rootfs_config: *const user_rootfs_config,
```
liqinggd, line 573: `rootfs_config_ptr: *const user_rootfs_config`

src/libos/src/fs/syscalls.rs
```diff
@@ -570,20 +570,31 @@ pub async fn do_ioctl(fd: FileDesc, cmd: u32, argp: *mut u8) -> Result<isize> {
...
     } else {
         Some(unsafe { key_ptr.read() })
     };
-    if occlum_json_mac_ptr.is_null() {
-        return_errno!(EINVAL, "occlum_json_mac_ptr cannot be null");
+    // If user provided valid parameters, do runtime mount and boot
+    // Otherwise, do general mount and boot
+    if !rootfs_config.is_null() {
+        from_user::check_ptr(rootfs_config)?;
+        let rootfs_config = user_rootfs_config::from_raw_ptr(rootfs_config)?;
```
liqinggd, line 584: `from_raw_ptr` is redundant.
```
let rootfs_config = unsafe { *rootfs_config_ptr };
```

.github/workflows/demo_test.yml
qzheng527: Ok

src/libos/src/fs/syscalls.rs
qzheng527: Updated in the latest commit.

liqinggd:
LGTM

demos/runtime_boot/init/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "init"
+version = "0.0.1"
+authors = ["LI Qing geding.lq@antgroup.com"]
+edition = "2018"
```
volcano0dr, line 5: Please use `edition = "2021"`

src/libos/src/config.rs
```diff
@@ -465,4 +505,120 @@ struct InputConfigMountOptions {
...
+    #[serde(default)]
+    pub entry_points: Vec<String>,
+    #[serde(default)]
+    pub mount: Vec<InputConfigMount>,
+}
+
+#[repr(C)]
+#[derive(Debug, Copy, Clone)]
+#[allow(non_camel_case_types)]
+pub struct user_rootfs_config {
```
volcano0dr, line 537: For future extensions, I recommend adding the struct length field.

src/libos/src/config.rs
```diff
@@ -98,9 +101,9 @@ pub struct Config {
     pub resource_limits: ConfigResourceLimits,
     pub process: ConfigProcess,
     pub env: ConfigEnv,
-    pub entry_points: Vec<PathBuf>,
+
```
volcano0dr, line 104: Please remove blank line.

src/libos/src/config.rs
```diff
@@ -216,22 +225,26 @@ impl Config {
...
-            }
-            mount
-        };
+
         Ok(Config {
             resource_limits,
             process,
             env,
-            entry_points,
+
```
volcano0dr, line 233: Please remove blank line

src/libos/src/config.rs
```diff
@@ -465,4 +516,146 @@ struct InputConfigMountOptions {
...
+impl ConfigApp {
+    pub fn from_user(config: &user_rootfs_config) -> Result<ConfigApp> {
+        // Check config struct length for future possible extension
+        if config.len != size_of::<user_rootfs_config>() {
+            return_errno!(EINVAL, "User Config Struct length not match");
+        }
+
+        // Update INIT_ENVS if necessary
+        if !config.envp.is_null() {
```
volcano0dr, line 580: I think it is possible to merge `init env` and `default env` to ConfigEnv, and rename ConfigEnv.default to trusted, what do you think?

qzheng527: Rename "default" to “trusted" makes sense. But this "default" corresponds to the name in Occlum.json. To make two consistent, it would be better to rename them together. While this change may require many existed demos change accordingly as well which is not so friendly to the Occlum users. So I think we can keep the "default" name as it is.

volcano0dr (approved):
LGTM

PR #337, opened 2022-11-24
Occlum yaml configuration

qzheng527:


etc/template/Occlum.yaml
```diff
@@ -0,0 +1,114 @@
...
+  #   type: hostfs
+  #   source: .
+  #
+  # Async FS mount
+  # - target: /sfs
+  #   type: async_sfs
+  #   source: ./run/async_sfs_image
+  #   options:
+  #     total_size: 4GB
+
```
volcano0dr, line 114: I think “untrusted_unix_socks” should be added.

volcano0dr (approved):
LGTM

Issue #328, opened 2022-11-18
[RFC] Introduce SwornDisk in NGO


lucassong-mh:
- Feature Name: Introduce SwornDisk
- Start Date: 2022-11-17

## In a nutshell

This RFC issue consists of two parts: One is SwornDisk high-level design overview, which explains SwornDisk's "why, how, what". One is [SwornDisk-Occlum](https://github.com/occlum/kxdisk-occlum)'s implementation review, which explains code structure and details of this SGX version.

# Design Overview
SwornDisk: A Log-Structured Secure Block Device for TEEs

Objectives: confidentiality, integrity, freshness, anonymity, consistency, and (flush) atomicity

![1](https://user-images.githubusercontent.com/22837133/202613207-907cd9ed-338b-47a4-a558-b3713e531faf.png)

## Motivation
- Existing solutions for protecting the on-disk data for TEEs are far from satisfactory in terms of both
  security and performance (eCryptfs, fscrypt, dm-crypt, SGX-PFS)
- Yet [SGX-PFS](https://www.intel.com/content/dam/develop/external/us/en/documents/overviewofintelprotectedfilesystemlibrary.pdf) has both performance issue (Slow random writes due to `2 × H` write amplification) and security vulnerability (Unanticipated snapshot attacks, [CVE-2022-27499](http://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2022-27499))
- Unanticipated snapshot attack: The adversary can capture and replay transient on-disk states (due to cache eviction in TEE) which are un-aware to users

![2](https://user-images.githubusercontent.com/22837133/202613234-30f5ecee-dc1a-498d-a90d-b83ddc169710.png)

## Background knowledge
1. In-place updates MHT-based approach VS. out-of-place updates log-structured approach

![3](https://user-images.githubusercontent.com/22837133/202613279-48610b2c-b29a-43d7-94ee-ea665c1f834f.png)

- Random writes are slower than sequential writes in HDD/SSD

- Write amplification: `2 × H` vs. `1 + ϵ (ϵ ≪ 1)`

2. Log-Structured Merge tree (LSM-tree)

- BigTable, Hbase, LevelDB, RocksDB

![4](https://user-images.githubusercontent.com/22837133/202613312-fcea2a91-2d0b-4f9e-a09a-69dad1de0c67.png)

- Disk-based index structure  for KV stores
- Workflow: KV pair → MemTable → Sorted String Table → Minor compaction to L0 → Major compaction to Li

[...]

PR #342, opened 2022-12-02
[crates] Add sworn-disk crate

lucassong-mh:
This PR adds `sworn-disk` to NGO's crates

RFC issue: https://github.com/occlum/ngo/issues/328


src/libos/crates/sworn-disk/src/lib.rs
```diff
@@ -0,0 +1,106 @@
+//! This crate provides the abstractions for SwornDisk.
+//!
+//! SwornDisk is a log-structured secure block device for TEEs, allowing file systems to
+//! be stacked upon it for transparent I/O protection. This crate is an Intel SGX version for Occlum
+//! where SwornDisk serves as a trusted logical block device within LibOS.
+//!
+//! Guarantees: confidentiality, integrity, freshness, anonymity, consistency, (flush) atomicity.
+//!
+//! SwornDisk uses a secure data log to organize encrypted data blocks into segments (out-of-place updates).
+//! It keeps the mapping between user-query block address (LBA) and eventually-persist block address (HBA) in TEE.
```
tatetian: Better stick with the definitions of Logical Block Addresss (LBA) and Host Block Addresss (HBA).

src/libos/crates/sworn-disk/src/lib.rs
```diff
@@ -0,0 +1,106 @@
...
+//! let root_key = DefaultCryptor::gen_random_key();
+//!
+//! // Create SwornDisk
+//! let sworn_disk = SwornDisk::create_new(Arc::new(sync_disk), root_key);
+//!
+//! let total_bytes = sworn_disk.total_bytes();
+//! let content = 5u8;
+//!
+//! // Write contents to SwornDisk
+//! let wbuf = [content; total_bytes];
```
tatetian: This example actually shows an anti-pattern: it allocates a 1GB array on the stack.

src/libos/crates/sworn-disk/src/lib.rs
```diff
@@ -0,0 +1,106 @@
...
+mod data;
+mod index;
+mod journal;
+mod prelude;
+mod superblock;
+mod sworn_disk;
+mod util;
+
+use self::checkpoint::Checkpoint;
+pub use self::config::{GiB, KiB, MiB, HBA, LBA, SEGMENT_SIZE};
```
tatetian: The Rust naming conventions is to name abbreviations like HBA, LBA, XML, CPU, etc as `Hba`, `Lba`, `Xml`, and `Cpu`. There are many such naming examples in std, e.g., `Utf8Chunk` (not `UTF8Chunk`), `IpAddr` (not `IPAddr`).

src/libos/crates/sworn-disk/src/superblock.rs
```diff
@@ -0,0 +1,181 @@
...
+use crate::checkpoint::{KeyTable, BITC, DST, RIT, SVT};
+use crate::prelude::*;
+
+use std::convert::TryInto;
+
+/// Superblock.
+#[derive(Debug, Clone)]
+pub struct SuperBlock {
+    /// Magic number
+    pub magic_number: usize,
```
tatetian: Use `u32` or `u64` instead of `usize` as the magic number has nothing to do with addresses or lengths.

src/libos/crates/sworn-disk/src/superblock.rs
```diff
@@ -0,0 +1,181 @@
...
+    /// Per segment size
+    pub segment_size: usize,
+
+    /// Number of data segments
+    pub num_data_segments: usize,
+    /// Number of index segments
+    pub num_index_segments: usize,
+
+    /// Address of the superblock region
+    pub superblock_region: HBA,
```
tatetian: `superblock_addr` or `superblock_hba` would make a better name. Using a name containing "region" like `superblock_region` would give the false impression that the field includes both the start and end addresses of the region.

src/libos/crates/sworn-disk/src/checkpoint/mod.rs
```diff
@@ -0,0 +1,180 @@
...
+    pub fn key_table(&self) -> &KeyTable {
+        &self.key_table
+    }
+}
+
+// TODO: Try use `serde` to serialize `Checkpoint`
+impl Checkpoint {
+    pub async fn persist(
+        &self,
+        disk: Arc<dyn BlockDevice>,
```
tatetian: There are as many as 22 locations that take `disk: Arc<dyn BlockDevice>` as a function argument. I think for most of them it makes more sense to keep an `Arc<dyn BlockDevice>` as a field of the struct. This makes the method interfaces simpler and avoid cloning every time the methods are called.

src/libos/crates/sworn-disk/src/util/lba_range.rs
```diff
@@ -0,0 +1,105 @@
+//! A range between two lbas.
+use crate::LBA;
+use block_device::BLOCK_SIZE;
+
+/// LBA range.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct LBARange {
+    pub first: LBA,
+    pub last: LBA,
```
tatetian: When deciding whether the fields of a struct should be public or not, the default answer should be NO. To my surprise, this crate seems to be more inclined to have public fields. The downside of having public fields is breaking encapsulation and leaking implementation details, which is against the best SW engineering practices.

In Rust, specifically, if a struct has all fields public, then the user can skip the constructor methods and create instances with whatever fields of his choice. In almost of cases, this is a bad idea because it makes it impossible for the type keep any invariances.

Take `LBARange` as an example. Apparently, all methods of `LBARange` assume `first <= last`. But this assumption is quite fragile as the users of `LBARange` can easily create `LBARange` instances that break the invariance.

Please review all type definitions and see if it is absolutely necessary to make their fields public.

src/libos/crates/sworn-disk/src/config.rs
```diff
@@ -0,0 +1,149 @@
...
+// Batch read threshold (Range query trigger condition)
+pub const BATCH_READ_THRESHOLD: usize = BLOCK_SIZE * 4;
+
+/// # Data
+// Segment Id
+#[allow(unused)]
+pub type SegmentId = usize;
+
+// Segment size (1024 blocks)
+pub const SEGMENT_SIZE: usize = MiB * 4;
```
tatetian: `4 * MiB` is more readable as it is aligned with the natural language.

Please review all such occurrences and revise accordingly.

src/libos/crates/sworn-disk/src/config.rs
```diff
@@ -0,0 +1,149 @@
...
+pub const ROOT_RECORD_SIZE: usize = size_of::<RootRecord>();
+pub const LEAF_RECORD_SIZE: usize = size_of::<LeafRecord>();
+
+/// # Cryption
+pub const CIPHER_SIZE: usize = BLOCK_SIZE;
+pub const AUTH_ENC_KEY_SIZE: usize = 16;
+pub const AUTH_ENC_MAC_SIZE: usize = 16;
+pub const AUTH_ENC_IV_SIZE: usize = 12;
+
+pub type KeyType = [u8; AUTH_ENC_KEY_SIZE];
```
tatetian: Why not simply call them `Key`, `Mac`, and `Iv`?

In addition, it is a surprise to me that these types are not defined as `util/cryption.rs`.

Personally, I find this config module does more harm than good. Because it exposes global parameters to every other parts of the crate. At least some of these configurations could be defined under their owner modules, close to the code that uses them and hidden from other modules that should not be concerned.

src/libos/crates/sworn-disk/src/config.rs
```diff
@@ -0,0 +1,149 @@
...
+pub const BA_SIZE: usize = size_of::<LBA>();
+pub const USIZE_SIZE: usize = size_of::<usize>();
+pub const U32_SIZE: usize = size_of::<u32>();
+pub const BITMAP_UNIT: usize = 8;
+
+// Test-purpose
+lazy_static::lazy_static! {
+    pub(crate) static ref OPEN_FAILED: AtomicBool = AtomicBool::new(false);
+}
+
```
tatetian: I don't think these two utility functions are qualified as configurations.

src/libos/crates/sworn-disk/src/config.rs
```diff
@@ -0,0 +1,149 @@
...
+pub const DATA_SIZE_PER_BIT: usize = MiB * 512;
+pub const MAX_LEVEL0_BIT_NUM: usize = 1;
+
+/// ## BIT
+// Unique id of a BIT (equals to its start hba)
+pub type BitId = HBA;
+
+// Level of a BIT
+#[allow(unused)]
+pub const MAX_BIT_LEVEL: u8 = 3;
```
tatetian: Is this parameter really configurable? The BIT implementation seems to be hardcoded to have three levels. I think this constant should be put under `index/bit.rs`. You see. It is much more likely for a parameter to be respected by the implementation if the parameter is kept close with the implementation, rather than away from it.

And by the way, who cares about the BIT levels instead of BITs themselves? Nobody. Not even the BITs. This parameter is never used...

src/libos/crates/sworn-disk/src/index/bit.rs
```diff
@@ -0,0 +1,591 @@
...
+//! Responsible for doing disk management, query processing, and security protection.
+use super::cache::BITCache;
+use super::metadata::SearchFlag;
+use super::record::{InternalRecord, LeafRecord, Record, RootRecord};
+use crate::prelude::*;
+use crate::util::RangeQueryCtx;
+
+use std::convert::TryInto;
+use std::fmt::{self, Debug};
+
```
tatetian: I think this module can use better encapsulation.

The `bit.rs`, `cache.rs`, `metadata.rs`, and `bit_builder.rs` files are for BITs. They can be put under a `bit/` directory, only expose the minimal set of APIs, and hide implementation details from the rest of the index module.

For example, I don't see why BIT cache is exposed to the user of BIT through `SearchFlag` (see my comments about `SearchFlag` below). Caching should be an performance optimization trick that is internal to a system (here, BITs). 

Another example is the internal BIT data structures like `RootBlock`, `InternalBlock`, and `LeafBlock`. Should they be public types that are visible to the rest of the index module? I don't think so.

src/libos/crates/sworn-disk/src/index/bit.rs
```diff
@@ -0,0 +1,591 @@
...
+    pub fn addr(&self) -> HBA {
+        self.id as _
+    }
+
+    /// Search the target record in BIT given a lba and a flag.
+    pub async fn search(
+        &self,
+        target_lba: LBA,
+        disk: Arc<dyn BlockDevice>,
+        flag: SearchFlag,
```
tatetian: I don't get the point of this search flag: why using cache is made an option? Isn't using cache always preferred?

src/libos/crates/sworn-disk/src/index/bit.rs
```diff
@@ -0,0 +1,591 @@
...
+                .search(*target_lba, disk.clone(), search_flag.clone())
+                .await
+            {
+                searched_records.push(record);
+                *completed = true;
+            }
+        }
+    }
+
+    /// Get all records from BIT.
```
tatetian: This method is not efficient. It copies records to a temp heap buffer for every compaction. Since BITs are immutable, it is safe to access BIT records without any copying. We can implement an iterator for this purpose.

src/libos/crates/sworn-disk/src/index/compaction.rs
```diff
@@ -0,0 +1,214 @@
...
+                checkpoint.clone(),
+            )
+            .await?;
+        }
+
+        // Pick an available segment from `index SVT`
+        let seg_addr = checkpoint.index_svt().write().pick_avail_seg().unwrap();
+
+        // TODO: Optimize this padding logic
+        Record::padding_records(&mut records, MAX_MEM_TABLE_CAPACITY);
```
tatetian: What is the point of padding?

src/libos/crates/sworn-disk/src/index/mem_table.rs
```diff
@@ -0,0 +1,332 @@
...
+use crate::data::CacheState;
+use crate::prelude::*;
+use crate::util::RangeQueryCtx;
+use crate::{Checkpoint, Record};
+
+use std::collections::BTreeMap;
+use std::ops::Bound::Included;
+
+/// MemTable. Define common memtable ability.
+pub trait MemTable<K: Into<LBA>, V: Clone> {
```
tatetian: This trait along with its generic parameters does not seem to be useful. There is only one implementation for the trait with fixed `K` and `V`. And we will always want to have just one single most efficient MemTable implementation. So defining the trait looks like a useless distraction to me.

src/libos/crates/sworn-disk/src/index/mem_table.rs
```diff
@@ -0,0 +1,332 @@
...
+    Negative(LBA),
+    /// Same lba, double records (one is regular, the other is negative)
+    RegularAndNegative(Record),
+}
+
+impl MemTable<LBA, MemRecord> for BTreeMemTable {
+    fn get(&self, key: &LBA) -> Option<MemRecord> {
+        self.records.get(key).map(|record| record.clone())
+    }
+
```
tatetian: This method only accepts `Regular` and `Negative`, not `RegularNegative`. But this restriction is not enforced at the type level. One way to make the API less error prone is to have two separate methods:

```
fn put(&mut self, key: LBA, record: Record)
```
    
and

```
fn remove(&mut self, key: LBA)
```

The new interface not only makes it impossible to feed invalid inputs, it can also hide `MemRecord`, an implementation-specific detail, from the user.

src/libos/crates/sworn-disk/src/index/mem_table.rs
```diff
@@ -0,0 +1,332 @@
...
+        }
+    }
+
+    /// Put a `MemRecord` into MemTable.
+    pub async fn put(
+        &self,
+        lba: LBA,
+        mem_record: MemRecord,
+        checkpoint: Arc<Checkpoint>,
+    ) -> Result<bool> {
```
tatetian: It seems to me that there is an easier way to enable the concurrency between the minor compaction and querying of an immutable MemTable.

Currently, we are using `CacheState` and `Pollee` and exposing all the state manipulation and wait methods (e.g., `set_state` and `wait_events`). Having many states and leaking these states to users (e.g., compactor) makes the code difficult to understand and reason.

Consider this alternative solution. The observation is that there seems to be only one boolean state that we really about, which is whether the MemTable is immutable or not. During the immutable period, the MemTable can be queried, but not updated. If a task attempts to update an immutable MemTable, then it should block. All other states like `Vacant`, `Full`, `Flushing`, and `Clearing` are not necessary.

The struct is simplified to the following form.

```rust
pub struct LockedMemTable { 
        mem_table: AsyncRwLock<Box<dyn MemTable<LBA, MemRecord>>>,
} 

impl LockedMemTable {
        /// Make the MemTable immutable, blocking any attempts to update.
        pub async fn freeze(&self) -> FreezeGuard<'_> {
            FreezeGuard(self.mem_table.read().await)
        }

        pub async fn put(&self, ...) -> Result<bool>  {
            let mut mem_table = self.mem_table.write().await;
            ...
        }
            
        pub async fn search(&self, ...) -> Option<Record> {
            let mem_table = self.mem_table.read().await;
            ... 
        }
}

pub struct FreezeGuard<'a>(AsyncRwLockGuard<'a, Box<dyn MemTable<LBA, MemRecord>>>);
```

This alternative approach not only simplifies MemTable, but also Compactor.

src/libos/crates/sworn-disk/src/index/mem_table.rs
```diff
@@ -0,0 +1,332 @@
...
+                self.records
+                    .insert(key, MemRecord::RegularAndNegative(r1.clone()));
+            }
+            (MemRecord::Regular(r1), Some(MemRecord::RegularAndNegative(r2))) => {
+                self.size -= 1;
+                let _ = reclaimed_hba.insert(r2.hba);
+                self.records
+                    .insert(key, MemRecord::RegularAndNegative(r1.clone()));
+            }
+            (MemRecord::Negative(_), Some(MemRecord::Regular(_))) => {
```
tatetian: Why no reclaim? And why not simply remove the key?

src/libos/crates/sworn-disk/src/index/mem_table.rs
```diff
@@ -0,0 +1,332 @@
...
+                self.records.insert(key, value.clone());
+            }
+            (MemRecord::Negative(_), Some(MemRecord::RegularAndNegative(_))) => {
```
tatetian: Why not reclaim?

src/libos/crates/sworn-disk/src/index/mem_table.rs
```diff
@@ -0,0 +1,332 @@
...
+            }
+            (_, None) => {
+                self.records.insert(key, value.clone());
+            }
+            (_, _) => {
+                panic!("This should not happen!");
+            }
+        }
+
+        self.size += 1;
```
tatetian: I don't quite understand what exactly does `size` count. Could you clarify the counting criteria?

src/libos/crates/sworn-disk/src/index/mem_table.rs
```diff
@@ -0,0 +1,332 @@
...
+        reclaimed_hba
+    }
+
+    fn get_range(&self, first: &LBA, last: &LBA) -> Vec<MemRecord> {
```
tatetian: It is more efficient to return an iterator, rather than a Vec.

Another issue is the name. In Rust, the convention for getter methods is not to have the `get_` prefix. Please fix all getter names. So this method is better called `records_in_range` or something simillar.

src/libos/crates/sworn-disk/src/index/mem_table.rs
```diff
@@ -0,0 +1,332 @@
...
+        self.records
+            .range((Included(first), Included(last)))
+            .map(|(_, mem_record)| mem_record.clone())
+            .collect()
+    }
+
+    fn get_all_records(&self) -> Vec<Record> {
```
tatetian: Same problem with `get_range`.

src/libos/crates/sworn-disk/src/index/metadata.rs
```diff
@@ -0,0 +1,110 @@
+//! Metadata for BITs of Lsm Tree.
+use super::bit::BIT;
+use super::cache::BITCache;
+use crate::prelude::*;
+
+use std::fmt::{self, Debug};
+
+/// Metadata of a BIT.
+/// It manages the underlying `BIT` and `BITCache`.
```
tatetian: `BITMeta` does not seem to be a good name. Because what it really contains is the BIT itself and its cache, which means `BITMeta` is conceptually equivalent to a BIT. So I think `BITMeta` should be named `Bit` and the original BIT should be named `DiskBit`. And `DiskBit` and `BitCache` should be hidden from BIT users. All the search methods should be implemented on `Bit`.

src/libos/crates/sworn-disk/src/index/record.rs
```diff
@@ -0,0 +1,164 @@
...
+    pub hba: HBA,
+    pub cipher_meta: CipherMeta,
+}
+
+/// Leaf node of BIT.
+#[derive(Clone, Debug, Hash, PartialEq, Eq)]
+pub struct LeafRecord(pub InternalRecord);
+
+impl Record {
+    pub fn build(lba: LBA, hba: HBA, cipher_meta: CipherMeta) -> Self {
```
tatetian: The Rust naming convention is `new`. Use the verb `build` when the code follows the builder pattern or it assembles an object in a non-trivial way.

src/libos/crates/sworn-disk/src/index/metadata.rs
```diff
@@ -0,0 +1,110 @@
...
+    bit: BIT,
+    cache: Arc<BITCache>,
+}
+
+impl BITMeta {
+    pub fn from(bit: BIT, cache: Arc<BITCache>) -> Self {
+        Self { bit, cache }
+    }
+
+    #[inline]
```
tatetian: The inline annotation is doing more harms than good. It is almost always better to let the compiler to decide whether to inline a function or not. The only exception is for crate-level public methods, in which case the compiler does not have all the knowledge about the call sites of the crate-level public methods. So the programmer is better positioned to make the right decision.

src/libos/crates/sworn-disk/src/index/record.rs
```diff
@@ -0,0 +1,164 @@
...
+    /// Group records' hbas in consecutive increasing order.
+    pub fn group_consecutive_records_by_hba(records: &[Record]) -> impl Iterator<Item = &[Record]> {
+        records.group_by(|r1, r2| r2.hba.saturating_sub(r1.hba) == BLOCK_SIZE)
+    }
+
+    pub fn padding_records(records: &mut Vec<Record>, align: usize) {
+        if records.is_empty() || records.len() == align {
+            return;
+        }
+        records.extend(vec![
```
tatetian: `Vec::resize_with` is more efficient as it avoids allocating another temp `Vec`.

src/libos/crates/sworn-disk/src/index/record.rs
```diff
@@ -0,0 +1,164 @@
...
+            hba: NEGATIVE_HBA,
+            cipher_meta: CipherMeta::new_uninit(),
+        }
+    }
+
+    pub fn is_negative(&self) -> bool {
+        self.hba == NEGATIVE_HBA
+    }
+
+    /// Group records' hbas in consecutive increasing order.
```
tatetian: This is a fragile API because it works on the assumption that the records have been sorted. The assumption is neither documented nor enforced by the type system. Since this API is only used once, we can let this single call site to define such a function

src/libos/crates/sworn-disk/src/sworn_disk.rs
```diff
@@ -0,0 +1,491 @@
...
+            cleaner,
+            root_key,
+        }
+    }
+
+    /// Open a created `SwornDisk` given the root key. (sync, used for `/dev/sworn_disk`)
+    ///
+    /// * `root_key` - The root key that encrypts basic structures of SwornDisk (e.g., SuperBlock and Checkpoint).
+    pub fn open(disk: Arc<dyn BlockDevice>, root_key: KeyType) -> Result<Self> {
+        if OPEN_FAILED.load(Ordering::SeqCst) {
```
tatetian: This looks weird. Even for testing purpose. Could you explain why keeping a global state is the absolutely necessary?

src/libos/crates/sworn-disk/src/sworn_disk.rs
```diff
@@ -0,0 +1,491 @@
...
+                root_key,
+            })
+        });
+        open_self
+    }
+
+    /// Open a created `SwornDisk` given the root key. (async, used by async filesystem)
+    ///
+    /// * `root_key` - The root key that encrypts basic structures of SwornDisk (e.g., SuperBlock and Checkpoint).
+    pub async fn open_with(disk: Arc<dyn BlockDevice>, root_key: KeyType) -> Result<Self> {
```
tatetian: Why we are having two open methods? `open` can be implemented explicitly by the user with `block_on` and `open_with`. It is not a good idea to use `block_on` internally because `block_on` has some serious limitations.

src/libos/crates/sworn-disk/src/sworn_disk.rs
```diff
@@ -0,0 +1,491 @@
...
+                self.checkpoint.clone(),
+            )
+            .await?;
+
+        Ok(buf.len())
+    }
+
+    /// Check if the arguments for a read or write is valid.
+    fn check_rw_args(&self, offset: usize, buf: &[u8]) {
+        assert!(
```
tatetian: Should not use assert and panic on failure for input errors.

src/libos/crates/sworn-disk/src/util/lba_range.rs
```diff
@@ -0,0 +1,105 @@
...
+        }
+
+        Some(LBARange {
+            first: self.first.min(rhs_lba_range.first),
+            last: self.last.min(rhs_lba_range.last),
+        })
+    }
+}
+
+impl Iterator for LBARange {
```
tatetian: It is not a good idea to make LBARange iterators. First, LBARange appears to be immutable. But the iterator interface breaks the immutability. Second, the iterator implementation breaks the invariance that `self.first <= self.last`, which is assumed by `num_covered_blocks`.

src/libos/crates/sworn-disk/src/util/serialize.rs
```diff
@@ -0,0 +1,204 @@
...
+use std::collections::HashMap;
+use std::convert::TryInto;
+
+/// Provide ability for encode/decode
+/// from bytes/type to type/bytes.
+///
+/// Used for exchanging data between memory and device.
+pub trait Serialize {
+    /// Encode from struct to bytes. (move)
+    fn encode(self) -> Vec<u8>
```
tatetian: Why we would like to have a default implementation that panics? If the trait has no default implementation, then the compiler would ensure that such an implementation is given for every `T: Serialize`. This is better than having the pancing implementation.

Another issue is the interface. Returning a `Vec` is a great overhead for simple types like `LBA`.

src/libos/crates/sworn-disk/src/util/serialize.rs
```diff
@@ -0,0 +1,204 @@
...
+    where
+        Self: Sized,
+    {
+        unimplemented!()
+    }
+
+    // Decode bytes to struct.
+    fn decode(buf: &[u8]) -> Self
```
tatetian: In general, decoding should be a fallable operation.

tatetian (commented):
I like the overall code structure; it is clear and modular.

The following is some of my initial comments. Most of them is about the index. I will submit more comments for other modules later.

src/libos/crates/sworn-disk/src/checkpoint/rit.rs
```diff
@@ -0,0 +1,103 @@
...
+use std::convert::TryInto;
+use std::fmt::{self, Debug};
+
+/// Reverse Index Table.
+/// Manage reverse mapping (hba => lba).
+pub struct RIT(HashMap<HBA, LBA>);
+
+// TODO: Make RIT an on-disk map
+impl RIT {
+    pub fn init() -> Self {
```
tatetian: Most, if not all, checkpoint data structures (like `RIT`) should have their constructors named `new`, instead of `init`. As a Rust convention, a constructor named `new` gives you an "empty" object. And usually, a `new` constructor is cheap. In contrast, a constructor named `init` gives the impression that it involves complex and expensive initialization logic. But here this constructor is plain and simple. The heavylifting part of initialization is done in the `load` method.

src/libos/crates/sworn-disk/src/util/serialize.rs
```diff
@@ -0,0 +1,204 @@
...
+            fn encode_copy(&self) -> Vec<u8> {
+                unsafe {
+                    std::mem::transmute_copy::<$target_struct, [u8; $struct_size]>(self).to_vec()
+                }
+            }
+        }
+    };
+}
+
+/// Implement `persist()` and `load()` for Checkpoint structures.
```
tatetian: This macro should be implemented in the checkpoint module. When I skimmed through the code of checkpointing data structures, I was surprised to find no code about loading (e.g., `RIT::init` is empty) and persistence. Since this macro implements methods for checkpoint data structures, it is better off stay with the data structures.

src/libos/crates/sworn-disk/src/checkpoint/mod.rs
```diff
@@ -0,0 +1,180 @@
...
+    bitc: RwLock<BITC>,
+    data_svt: RwLock<SVT>,
+    index_svt: RwLock<SVT>,
+    dst: RwLock<DST>,
+    rit: RwLock<RIT>,
+    key_table: KeyTable,
+}
+
+// TODO: Maintain two copies of checkpoint for journal
+// TODO: Introduce incremental disk update for better performance
```
tatetian:
```
// TODO: Support on-demand loading, rather than loading the entire 
// checkpoint data structures all at once.
```

src/libos/crates/sworn-disk/src/checkpoint/key_table.rs
```diff
@@ -0,0 +1,120 @@
+//! Cryption Key Table for data segments.
+use crate::prelude::*;
+
+use std::collections::HashMap;
+use std::convert::TryInto;
+use std::fmt::{self, Debug};
+
+/// Cryption key table.
+/// Manage per-segment cryption keys.
```
tatetian: This `KeyTable` data structure that only manages per-segment keys cannot support writing threaded logging.

src/libos/crates/sworn-disk/src/index/lsm_tree.rs
```diff
@@ -0,0 +1,319 @@
...
+        let is_full = mem_table.put(lba, record, checkpoint.clone()).await?;
+
+        // If MemTable is full, trigger minor compaction
+        // Immutable MemTable -> BIT
+        if is_full {
+            // Turn to immutable
+            self.immut_idx.store(
+                self.immut_idx.load(Ordering::Relaxed) ^ 1,
+                Ordering::Relaxed,
+            );
```
tatetian: This one is simpler, more readable, and has the advantage of being atomic. 

```
self.immut_idx.fetch_xor(1, Ordering::Relaxed);
```

Using a load plus a store is concurrently safe only because there is a lock acquired by the caller of this method, making it less obvious why the code is correct.

I am not sure whether the atomic xor is more efficient than atomic load plus store. But here efficiency is of a lessor concern as filling up a MemTable is orders of magnitude more expensive than an atomic xor.

src/libos/crates/sworn-disk/src/index/bit_builder.rs
```diff
@@ -0,0 +1,209 @@
...
+        let leaf_records = self.build_leafs(records);
+
+        // Build internal blocks and construct internal records
+        let internal_records = self.build_internals(&leaf_records);
+
+        // Build root block and construct a BIT
+        let bit = self.build_bit(&internal_records, level, version);
+
+        // Write back BIT content to disk
+        disk.write(self.addr, self.buf.as_slice()).await?;
```
tatetian: I am uncomfortable with writing disk this way because the code is vulnerable to "on-disk buffer overflow".

Initially, `BITBuilder` is only given an address; later, the `build` method is called given an arbitrary number of records. Judging from the code of BIT builder _alone_, there is no way to tell if this line of `disk.write` would (mistakenly) overwrite other on-disk data structures. Whether `BITBuilder` would cause on-disk buffer overflow depends on whether the user follows some unwritten contract. This makes it harder to reason about the correctness of the code.

My suggestion is that the BIT builder does not take an address, but allocates segments on demand as it writes the BIT to the disk. This way, we can guarantee that there is enough space. Besides, it may benefit to introduce a concept called `DiskSlice`, which is similar to `slice`, but for disk. Writing to `DiskSlice` rather than writing to `dyn BlockDevice` prevents any unintended overwrites.

src/libos/crates/sworn-disk/src/index/compaction.rs
```diff
@@ -0,0 +1,214 @@
...
+            self.exec_major_compaction(
+                l0_bit_meta.as_ref().unwrap().clone(),
+                disk.clone(),
+                checkpoint.clone(),
+            )
+            .await?;
+        }
+
+        // Pick an available segment from `index SVT`
+        let seg_addr = checkpoint.index_svt().write().pick_avail_seg().unwrap();
```
tatetian: So a BIT can always fit into a segment? How realistic is this assumption?

In the current implementation, a BIT must fit into a segment and is limited to three levels. Furthermore, the LSM-tree is limited to two levels. This means that there will be an unbounded number of BITs at L1 as the total size of the index increases. How this configuration would impact the performance?

src/libos/crates/sworn-disk/src/index/lsm_tree.rs
```diff
@@ -0,0 +1,319 @@
...
+        }
+        mem_table.set_state(CacheState::Full);
+        self.compactor
+            .exec_minor_compaction(mem_table, disk, checkpoint)
+            .await
+    }
+}
+
+impl Debug for LsmTree {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
```
tatetian: Try not to use non-standard debug format for Rust objects (e.g., `"===LsmTree==="`). For structs like LsmTree, it is usually a good idea to use [`DebugStruct`](https://doc.rust-lang.org/std/fmt/struct.DebugStruct.html).

tatetian (commented):
More comments.

tatetian (commented):
What I am about to say does not mean I am encouraging the PR author to take the advice into action now. Because if it was to be done, it would mean a major overhaul. But I think it is beneficial to discuss the future direction of improvement now.

Personally, I think the current codebase suffers a structural problem. The problem is that the codebase looks modular, but the modules are in fact highly entangled.

One of the main symptom is that the index cannot be tested alone. The index directly depends on the checkpoint. Without the checkpoint initialized, it is impossible to do anything on the index. The checkpoint not only contains the metadata required by the index, but also the rest of SwornDisk. So it is safe to say that the index code is entangled with the rest of SwornDisk. 

I think that the current codebase is under too much influence of the paper's design. The paper's design is mainly concerned with security and clarity, not software engineering practices. So making the codebase closely follow the paper's design is harmful to the codebase.

But no worry. We have the time to figure out a plan to refactor the code structure later.

src/libos/crates/sworn-disk/src/index/mem_table.rs
lucassong-mh: Is this approach holding a guard for too long? Compaction is costy and a potential major compaction would occur before minor compaction.

src/libos/crates/sworn-disk/src/index/mem_table.rs
lucassong-mh: Under current implementation, A `Negative` is put into memtable only when GC started and check a HBA is still valid (no newer data).

`Regular` and `Negative` is easy to understand. But `RegularAndNegative` here always means the regular one is newer coming than the negative one.

We only reclaim a HBA when `Regular` meet a `Regular`.

When a `Negative` arrives, all older ones can be deprecated (they are already reclaimed and are now processed by GC)

The `Negative` ones cannot be removed directly in memtable since they are needed in Compaction.

Workflow in memtable:

| newer    | older              |                                                           |
| -------- | ------------------ | --------------------------------------------------------- |
| Regular  | Regular            | reclaim immediately and insert newer one                  |
| Regular  | Negative           | replace older one to `RegularAndNegative`                 |
| Regular  | RegularAndNegative | reclaim immediately                                       |
| Negative | Regular            | replace older one to newer one (only one `Negative` left) |
| Negative | RegularAndNegative | replace older one to newer one (only one `Negative` left) |
| _        | None               | insert newer one                                          |
| _        | _                  | should not happen                                         |

And I should add this to comment.

src/libos/crates/sworn-disk/src/index/compaction.rs
lucassong-mh: Under current implementation, the size of an on-disk BIT is settled (`INDEX_SEGMENT_SIZE`), it always includes a fixed-size records (since I use fixed-size array rather vector inside BIT nodes) and always manages same amount data (512MB in this case, can be configured).

So the max number of BITs in L1 is calculated at first (always manage all data region), so is index_svt's capacity, BIT start addresses and index region.

So I would need a padding logic in compaction and an exact counter in memtable. And a `BITBuf` has known-size content and can always write to disk successfully.

src/libos/crates/sworn-disk/src/index/mem_table.rs
lucassong-mh: `RegularAndNegative` now would occupy two slots in memtable. We need exact capacity info to decide whether memtable is full and start doing compaction.

See detailed response in https://github.com/occlum/ngo/pull/342#discussion_r1057221857

src/libos/crates/sworn-disk/src/index/cache.rs
```diff
@@ -0,0 +1,144 @@
...
+use crate::prelude::*;
+use lru::LruCache;
+
+/// A cache for BIT. It caches tree structure in memory.
+/// The root and internal nodes are all cached, leaf nodes is LRU cached.
+pub struct BITCache {
+    root_block: RwLock<Option<RootBlock>>,
+    internal_blocks: RwLock<Vec<InternalBlock>>,
+    leaf_capacity: usize,
+    lru_leaf_blocks: RwLock<LruCache<LeafRecord, Box<LeafBlock>>>,
```
tatetian: Keep `Arc<LeafBlock>` instead of `Box<_>`.

src/libos/crates/sworn-disk/src/checkpoint/key_table.rs
lucassong-mh: mark with `TODO`

src/libos/crates/sworn-disk/src/util/disk_view.rs
```diff
@@ -0,0 +1,79 @@
...
+use crate::prelude::*;
+
+use std::fmt::{self, Debug};
+
+/// Disk view that manages underlying disk and
+/// restrict read/write boundaries.
+#[derive(Clone)]
+pub struct DiskView {
+    // [begin, end)
+    boundary: (Hba, Hba),
```
tatetian: Consider using `core::ops::Range<Hba>`, which is more fitting than `(Hba, Hba)` to represent a range. For one thing, you don't have to add an extra comment to explain the start is inclusive and the end is exclusive.

src/libos/crates/sworn-disk/src/util/disk_view.rs
```diff
@@ -0,0 +1,79 @@
...
+    // Underlying disk
```
tatetian: We should resist the temptation to write superficial comments (like this one) and should focus on writing clear code and meaningful comments.

src/libos/crates/sworn-disk/src/util/disk_view.rs
```diff
@@ -0,0 +1,79 @@
...
+    disk: Arc<dyn BlockDevice>,
+}
+
+impl DiskView {
+    pub fn new(boundary: (Hba, Hba), disk: Arc<dyn BlockDevice>) -> Self {
+        debug_assert!(
+            boundary.0 % BLOCK_SIZE == 0
+                && boundary.1 % BLOCK_SIZE == 0
```
tatetian: I thought LBAs and HBAs are block IDs, not offsets in bytes, am I wrong? If LBAs and HBAs were offset in bytes, why would the index take `LBA` as keys?

src/libos/crates/sworn-disk/src/util/disk_view.rs
```diff
@@ -0,0 +1,79 @@
...
+                && boundary.0 <= disk.total_bytes()
+                && boundary.1 <= disk.total_bytes(),
```
tatetian: No check for `boundary.0 <= boundary.1`?

src/libos/crates/sworn-disk/src/util/disk_array.rs
```diff
@@ -0,0 +1,76 @@
+//! Disk array.
+use crate::prelude::*;
+
+use std::collections::BTreeMap;
+use std::fmt::Debug;
+use std::iter::FromIterator;
+use std::mem::size_of;
+
+/// Disk array that manages on-disk structures.
+pub struct DiskArray<T: Sized + Clone + Serialize + Debug> {
```
tatetian: One tips for writing trait bounds: _add the minimum trait bounds required for the specific context (e.g., impl blocks, function bodies, and type definitions)_. For example, `T: Clone` is only useful when implementing `Clone` for `DiskArray<T>`. `Sized` is a special trait bound that all `T` implicitly requires by default (if `Sized` is unwanted, give `?Sized` to explicitly). Thus, none of the trait bounds here are necessary or useful.

In almost all type definitions, we do not need trait bounds. Because almost all traits are only useful for implementation

src/libos/crates/sworn-disk/src/util/disk_array.rs
```diff
@@ -0,0 +1,76 @@
...
+    start_addr: Hba,
+    // k: offset(0, 1, ...), v: disk unit `T`
+    table: BTreeMap<usize, T>,
+    disk: DiskView,
+}
+
+impl<T: Sized + Clone + Serialize + Debug> DiskArray<T> {
```
tatetian: Again, it is recommended to keep only the necessary trait bounds. Here, this means simplifying the trait bounds to `T: Serialize`.

One common practice is to group impl blocks according to the required trait bounds.

```rust
impl<T> DiskArray<T> {
    pub fn new(start_addr: Hba, disk: DiskView) -> Self { .. }

    pub fn set(&mut self, offset: usize, unit: T) -> Option<T> { ... }

    pub fn unit_size() -> usize { ... }

    pub fn table_size(&self) -> usize { ... }
}

impl<T: Serialize> DiskArray<T> {
    pub async fn persist(&self, _root_key: &Key) -> Result<()> { ... }

    pub async fn get(&mut self, offset: usize) -> Option<&T> { ... }
}

impl<T: Debug> Debug for DiskArray<T> { ... }
```

src/libos/crates/sworn-disk/src/util/lba_range.rs
```diff
@@ -0,0 +1,121 @@
+//! A range between two lbas.
+use crate::Lba;
+use block_device::BLOCK_SIZE;
+
+/// Lba range.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct LbaRange {
+    first: Lba,
+    last: Lba,
+}
```
tatetian: It seems that a type for HBA ranges is also useful (consider `DiskArray::new(hba_range, ..)`.

```rust
pub struct DiskRange<B> {
    first: B,
    last: B,
}

pub type LbaRange = DiskRange<Lba>;
pub type HbaRange = DiskRange<Hba>;
```

But if we have a `DiskRange<T>`, what makes it different from `core::ops::Range<T>`? 

The first answer that comes across to me is the utility methods like `is_within_range`. But they can be implemented as trait extensions to `Range<T>`.

```rust
pub trait RangeExts<T> {
    fn is_sub_range(&self, rhs: &Range<T>) -> bool;
    fn is_overlapped(&self, rhs: &Range<T>) -> bool;
}

impl RangeExts<T> for Range<T> {
    fn is_sub_range(&self, rhs: &Range<T>) -> bool { ... }
    fn is_overlapped(&self, rhs: &Range<T>) -> bool { ... }
}
```

So this answer is not a strong argument to having a new type replicating what `Range<T>` has done.

After having a closer look at code, my conclusion is that `DiskRange<T>` has only one advantage over `Range<T>`, which is its awareness of the block alignment. See the iterator code and the `check_legal_range` method.

But is this really an advantage or a coverup for a deeper problem, the problem that _the codebase uses byte offsets in places where block offsets should be used_. As a block device, SwornDisk has almost no reason to deal with byte offsets internally. If SwornDisk is implemented to deal with blocks only (not bytes), then there is no point i having `DiskRange<T>` that offers no additional values over `Range<T>` except bridging byte offsets to block offsets (as done in the iterators and `check_legal_range`).

src/libos/crates/sworn-disk/src/util/range_query_ctx.rs
```diff
@@ -0,0 +1,57 @@
...
+    pub fn iter_mut(&mut self) -> std::collections::btree_map::IterMut<'_, Lba, (usize, bool)> {
+        self.range_ctx.iter_mut()
+    }
+
+    pub fn target_range(&self) -> &LbaRange {
+        &self.lba_range
+    }
+
+    pub fn get_offset(&self, lba: &Lba) -> usize {
+        self.range_ctx.get(lba).unwrap().0
```
tatetian: Since a LBA can be mapped to an offset (within the buffer) trivially, I think we should avoid storing an `usize` value in the map of `BTreeMap<Lba, (usize, bool)>`.

In fact, I think using `BTreeMap` is an overkill here. As I understand, the LBAs covered by an instance of `RangeQueryCtx` must be consecutive. So a linear data structure like `Vec` should be sufficient.  And since we do not need to store offsets (relative to the buffer), a bitmap would be most suitable.

src/libos/crates/sworn-disk/src/util/serialize.rs
```diff
@@ -0,0 +1,181 @@
...
+    /// Encode from struct to bytes.
+    fn encode(&self, encoder: &mut impl Encoder) -> Result<()>;
+
+    /// Decode bytes to struct.
+    fn decode(buf: &[u8]) -> Result<Self>
+    where
+        Self: Sized;
+
+    /// Encoded length of bytes.
+    fn encoded_len(&self) -> usize {
```
tatetian: I think this interface needs polishing.

First of all, it deserves a better name. The word "encoded" seems to suggest an object `obj: dyn Serialize` is _already_ encoded somehow. But this is not true. The method returns the predicted length if the object is encoded. A better name would be "binary_len" or "fixed_len".

Second, this method should be marked _fallible_ in an explicit way. Because apparently, not all objects can predict its length after being encoded into its binary form. In Rust, we do not express the concept of "null" by repurposing some values (say, 0) with some meaning. This kind of design has bee proven to be error-prone. 

There are two options to express a method being optional in Rust. The first one is making the method returning an `Option<usize>` and having a default implementation of returning `None`. The second one is to define a sub-trait `FixedLengthSerialize: Serialize`, which requires a method named `fixed_len` to be implemented.

src/libos/crates/sworn-disk/src/index/lsm_tree.rs
```diff
@@ -0,0 +1,305 @@
...
+use crate::util::RangeQueryCtx;
+use crate::{Checkpoint, Record};
+
+use std::fmt::{self, Debug};
+use std::sync::atomic::{AtomicU8, Ordering};
+use std::sync::Arc;
+
+/// Log-structured merged tree.
+pub struct LsmTree {
+    addr: Hba,
```
tatetian: What is the purpose of this address? I can't find any code that refers to this field.

src/libos/crates/sworn-disk/src/index/lsm_tree.rs
```diff
@@ -0,0 +1,305 @@
...
+        let aw_lock = self.arw_lock.write().await;
+
+        self.insert_record(lba, record).await?;
+
+        drop(aw_lock);
+        Ok(())
+    }
+
+    /// Search a record from lsm tree.
+    pub async fn search(&self, target_lba: &Lba) -> Option<Record> {
```
tatetian: Be consistent with yourself. Both the `insert` and `search` methods take `Lba` as arguments. But one chooses to use `Lba`, while the other `&Lba`. It does't make much difference which to use here (in terms of efficiency and readability). But an author should make up his mind and stick with it.

Personally, I would prefer `Lba` over `&Lba` because 1) using a value is always more straightforward than a reference to it, and 2) the size of a LBA is equal to or shorter than that of a memory address on 64-bit machines.

src/libos/crates/sworn-disk/src/index/compaction.rs
```diff
@@ -0,0 +1,205 @@
...
+        immut_mem_table.clear().await;
+
+        drop(guard);
+        complete_fn();
+        Ok(())
+    }
+
+    /// Major compaction. (Li -> Li+1)
+    /// Currently L0 -> L1.
+    pub async fn exec_major_compaction(
```
tatetian: To avoid misuse from the LSM-tree, this `exec_major_compaction` method should be private. Or it should be protected with an async lock. This helps prevent concurrent major compactions (just like the minor ones).

src/libos/crates/sworn-disk/src/index/lsm_tree.rs
```diff
@@ -0,0 +1,305 @@
...
+        waiter.wait().await?;
+        self.wq.dequeue(&mut waiter);
+        Ok(())
+    }
+}
+
+impl LsmTree {
+    /// Persist MemTables to BIT.
+    pub async fn persist(&self) -> Result<()> {
+        let mem_table = self.mem_table();
```
tatetian: This method should acquire a write guard on `self.arw_lock` to prevent certain race conditions.

First of all, this method itself is not thread-safe: the `persist` method may be called from multiple concurrent thread and it does not make sense to have the current mutable MemTable to be dumped more than once.

Second, I am worried that a concurrent inserting thread may trigger the two MemTables to switch. In this case,  the `mem_table` may return either the immutable or mutable one. I am not entirely sure whether this is ok or not.

Furthermore, allowing minor compactions to be performed on both mutable and immutable looks a bad idea. Normally, minor compactions are performed on the immutable one (see `insert_record`). But this `persist` method triggers minor compactions on the mutable one. This strategy is harder to understand and, more importantly, may cause correctness problems: if the `persist` and `insert` methods are executing concurrently, then the immutable MemTable may be persisted _after_ the mutable one, causing newer records to be overridden by older ones. Note that acquiring a write guard on `self.arw_lock` alone cannot fix this problem.

src/libos/crates/sworn-disk/src/index/lsm_tree.rs
```diff
@@ -0,0 +1,305 @@
...
+
+    async fn wait_compaction(&self, mem_table: &Arc<LockedMemTable>) -> Result<()> {
+        if !mem_table.is_full() {
+            return Ok(());
+        }
+        let mut waiter = Waiter::new();
+        self.wq.enqueue(&mut waiter);
+        waiter.wait().await?;
+        self.wq.dequeue(&mut waiter);
+        Ok(())
```
tatetian: This method needs to be refactored as the following form to prevent losing wakeup events and sleeping forever. (Can you think of an example race condition?)

```rust
        let mut waiter = Waiter::new();
        self.wq.enqueue(&mut waiter);
        if !mem_table.is_full() {
            return Ok(());
        }
        waiter.wait().await?;
        self.wq.dequeue(&mut waiter);
```

To optimize its performance, we can improve it further.

```rust
       // Fast path
        if !mem_table.is_full() {
            return Ok(());
        }

        // Slow path
        let mut waiter = Waiter::new();
        self.wq.enqueue(&mut waiter);
        if !mem_table.is_full() {
            return Ok(());
        }
        waiter.wait().await?;
        self.wq.dequeue(&mut waiter);
```

Our last concern is that `Waiter::wait` may return due to TIRQs (Task Interrupt ReQusts)---and yes, it is documented with the `wait` method. To deal with these unwanted wakeups, we should add a loop to ensure that when this method returns, the MemTable is guaranteed to have free space.

```rust
        // Fast path
        if !mem_table.is_full() {
            return Ok(());
        }

        // Slow path
        let mut waiter = Waiter::new();
        self.wq.enqueue(&mut waiter);
        while !mem_table.is_full() {
            let _ = waiter.wait().await;
        }
        self.wq.dequeue(&mut waiter);
```

`Waiter` and `WaiterQueue` are low-level stuff. That's why they require more efforts to use correctly.

src/libos/crates/sworn-disk/src/index/lsm_tree.rs
```diff
@@ -0,0 +1,305 @@
...
+                .await?;
+        }
+
+        Ok(())
+    }
+
+    /// Concrete search logic.
+    /// MemTable -> Immutable MemTable -> L0 BIT -> L1 BITs
+    async fn search_record(&self, target_lba: &Lba) -> Option<Record> {
+        // Search for MemTable
```
tatetian: It is to search the target record _in_ MemTable, not for.

src/libos/crates/sworn-disk/src/index/mem_table.rs
```diff
@@ -0,0 +1,195 @@
...
+
+/// `Record` in memory.
+#[derive(Clone, Debug)]
+pub enum MemRecord {
+    /// A regular record
+    Regular(Record),
+    /// A negative record, only its lba is meaningful
+    Negative(Lba),
+    /// Same lba, double records (one is regular, the other is negative)
+    RegularAndNegative(Record),
```
tatetian: To make the intention crystal clear, I think a better name would be `NegativeThenRegular`.

src/libos/crates/sworn-disk/src/index/mem_table.rs
```diff
@@ -0,0 +1,195 @@
...
+    }
+
+    /// Put a record into MemTable.
+    pub async fn put<F, Fut>(
+        &self,
+        lba: Lba,
+        record: Record,
+        checkpoint: &Arc<Checkpoint>,
+        reclaim_fn: F,
+    ) -> Result<bool>
```
tatetian: The function signature can be simplified to:

```rust
    /// Put a record into MemTable.
    pub async fn put<F, Fut>(
        &self,
        lba: Lba,
        record: Record,
        reclaim_fn: F,
    ) -> Result<bool>
    where
        F: FnOnce(Record, MemRecord) -> Fut,
        Fut: Future<Output = MemRecord>,
```

Removing the checkpoint arguments makes the function signature looks more clean.

The caller should be rewritten accordingly:

```rust
    async fn insert_record(&self, lba: Lba, record: Record) -> Result<()> {
        ...

        let reclaim_fn = async |new, old| {
             apply_memtable_reclaim_policy(self.checkpoint, new, old).await
        };
        let is_full = mem_table.put(lba, record, reclaim_fn).await.unwrap();
```

The above code should work. But I haven't compiled the code myself so I can't make any promises. Rust's async/await is a bit tricky. If it was not working, we should be able to tweak the above code to make it work.

src/libos/crates/sworn-disk/src/index/reclaim.rs
```diff
@@ -0,0 +1,54 @@
...
+/// Currently, A `Negative` is put into memtable only when GC(garbage collection) started and check a Hba is still valid (no newer data).
+/// `Regular` and `Negative` is easy to understand. But `RegularAndNegative` here always means the regular one is newer coming than the negative one.
+/// We only reclaim a Hba when `Regular` meet a `Regular`.
+/// When a `Negative` arrives, all older ones can be deprecated (they are already reclaimed and are now processed by GC)
+/// The `Negative` ones cannot be removed directly in memtable since they are needed in compaction.
+pub async fn apply_memtable_reclaim_policy(
+    checkpoint: Arc<Checkpoint>,
+    coming_record: Record,
+    existed_record: MemRecord,
+) -> MemRecord {
```
tatetian: Pass references to avoid unnecessary memory copying.

src/libos/crates/sworn-disk/src/index/mem_table.rs
```diff
@@ -0,0 +1,195 @@
...
+            return_errno!(EINVAL, "memtable is full, not available to put");
+        }
+
+        let mut mem_table = self.mem_table.write().await;
+        let inc_num_records = || {
+            self.num_records.fetch_add(1, Ordering::Relaxed);
+        };
+        if let Some(existed_record) = mem_table.get(&lba) {
+            let to_be_replaced =
+                reclaim_fn(checkpoint.clone(), record, existed_record.clone()).await;
```
tatetian: When I initially proposed adding `reclaim_fn` as an input to MemTable, I meant registering a _callback_. One nice property about callbacks is that one can add empty callbacks or none callbacks at all. Callbacks are plugins to a framework. Their correctness won't affect that of the framework. And their complexity won't add any burden to a reader who tries to understand the framework.

But the current implementation does not treat `reclaim_fn` as a callback. Without a given callback or a working callback, the MemTable's `put` method cannot work. Looking at the code of `apply_memtable_reclaim_policy`, I believe a key part of this function should have been present in the `put` method, rather than `apply_memtable_reclaim_policy`.

Lastly, I am wondering why the callback function is passed at a per-call basis, instead of registered at a per-object basis (i.e., upon `MemTable` creation). The flexibility brought by per-call callbacks does not seem to be useful to us. So I would recommend choosing the per-object approach.

src/libos/crates/sworn-disk/README.md
```diff
@@ -0,0 +1,185 @@
+## JinDisk (SwornDisk)
```
tatetian: I appreciate the effort that you have been put into writing this README full of technical details. But I would highly suggest removing most of them. This is because this README reads like a short paper (or an extended abstract), but it should have been read like a project README. The two types of documents serve different audience, thus having different structures and focuses. A paper emphasizes technical novelties and intern workings (this is what the current README does). A project README focuses on selling the project to its (potential) users, explaining why it is useful and how to use. See the [README of Linux's version of JinDisk](https://github.com/jinzhao-dev/jindisk). And since this main user is Occlum, we don't even need to sell. In addition, we will soon make the research paper available on Arxiv.org, before its official publication. The paper should be the primary source for technical details. After the Arxiv version is available, we will add a link to the paper in this project README.

In short, keep this README short and concise and focus on providing information that is absent in the research paper.

tatetian (commented):
The new version looks promising. Good job on improving the code quality.

One major issue is that we need to rename SwornDisk to JinDisk. The crate name should be `jindisk` (Yes, no `-` between `jin` and `disk`) as JinDisk should be considered one name and word, not two. On the other hand, Jinzhao Disk is two words.

src/libos/crates/sworn-disk/src/index/reclaim.rs
lucassong-mh: Mark with `TODO`: Storing async closure as struct field requires `'static`

src/libos/crates/sworn-disk/src/checkpoint/key_table.rs
lucassong-mh: Can each threaded-logging block use corresponding per-segment keys?

src/libos/crates/block-device/src/block_id.rs
```diff
@@ -0,0 +1,84 @@
+//! Block id.
+use core::convert::TryInto;
+use core::ops::{Add, AddAssign, Sub, SubAssign};
+
+use crate::prelude::*;
+use errno::return_errno;
+
+type Id = usize;
+type Offset = usize;
```
tatetian: For the same reason as above, this type alias should be made public if it is really useful. But I don't think it is. I think everyone would simply use `usize`. So having this type alias won't do us any good.

src/libos/crates/block-device/src/block_id.rs
```diff
@@ -0,0 +1,84 @@
+//! Block id.
+use core::convert::TryInto;
+use core::ops::{Add, AddAssign, Sub, SubAssign};
+
+use crate::prelude::*;
+use errno::return_errno;
+
+type Id = usize;
```
tatetian: First of all, I think `u64` is preferable than `usize`, as explained in [a post on Rust user forum](https://users.rust-lang.org/t/i32-vs-isize-u32-vs-usize/22657/3).

> Use usize and isize when it's related to memory size -- the size of an object, or indexing a vector, for instance. It will be a 32-bit number on 32-bit platforms, as that's the limit of memory they can address, and likewise for 64-bit.
> 
> Use u32 and i32 when you just want numbers.

The issue has been brought up before.

image

Second, this type alias should be made public. Why would a _private_ type alias be used as part of the _public_ API? If the readability of this implementation (this file) can benefit from the type alias (I believe this is true), then that of the user (other files) should do, too. So it seems logical to make the type alias public.

```rust
pub type RawBlockId = u64;
```

src/libos/crates/block-device/src/block_id.rs
```diff
@@ -0,0 +1,84 @@
...
+impl BlockId {
+    pub const fn new(id: Id) -> Self {
+        debug_assert!(
+            BLOCK_SIZE.saturating_mul(id) <= usize::MAX as usize,
+            "block id is too big"
+        );
+        Self(id)
+    }
+
+    pub fn id(&self) -> Id {
```
tatetian: This signature is better:

```rust
pub fn to_raw(self) -> RawBlockId {
```

which would make the user code more readable.

Compare the user code using the new name

```rust
bid.to_raw()
```

with that using the old one

```rust
bid.id()
```

The former code makes it apparent that a block ID is turned into its raw form. The latter one looks non-sense or stupid because it looks like we are getting the ID of an ID.

src/libos/crates/block-device/src/block_id.rs
```diff
@@ -0,0 +1,84 @@
...
+        self.0
+    }
+
+    pub fn to_offset(&self) -> Offset {
```
tatetian: 1. It is more preferable to use `self` instead of `&self` in the methods of a type `T: Copy` whose size is no greater than the pointer size. Such examples are abundant in std, e.g., `u32`, `ptr`, `NonNull`. The rationale is that `self` is easier to work with than `&self`, and copying `self` is even cheaper than dereferencing `&self`.

2. And how about renaming this method to `to_byte_offset`? For most of the time we work with offsets in _bytes_. But sometimes we want to deal with offsets in _blocks_ or some other units. If the method name explicitly states the units (bytes), then there won't be any confusion.

src/libos/crates/block-device/src/block_id.rs
```diff
@@ -0,0 +1,84 @@
...
+
+    fn check_align(offset: Offset) -> Result<()> {
+        if offset % BLOCK_SIZE != 0 {
+            return_errno!(EINVAL, "offset not aligned with block size")
+        }
+        Ok(())
+    }
+}
+
+impl Add for BlockId {
```
tatetian: It does not make any sense to add two block IDs. We should implement `Add<U64>`. Other arithmetic operations have similar problems.

```rust
impl Add<u64> for BlockId {
   type Output = Self;

   ...
}
```

src/libos/crates/block-device/src/block_id.rs
```diff
@@ -0,0 +1,84 @@
...
+use core::ops::{Add, AddAssign, Sub, SubAssign};
+
+use crate::prelude::*;
+use errno::return_errno;
+
+type Id = usize;
+type Offset = usize;
+
+#[derive(Copy, Clone, Debug, Hash, PartialEq, Eq, PartialOrd, Ord)]
+pub struct BlockId(Id);
```
tatetian: How about renaming `BlockId` to `Bid`? It seems that we are going to use it throughout the codebase. So having a shorter name makes sense to me.

src/libos/crates/block-device/src/block_id.rs
```diff
@@ -0,0 +1,84 @@
...
+
+    pub fn id(&self) -> Id {
+        self.0
+    }
+
+    pub fn to_offset(&self) -> Offset {
+        self.0 * BLOCK_SIZE
+    }
+
+    pub fn from_offset(offset: Offset) -> Result<Self> {
```
tatetian: I am wondering whether this method should be fallible or not. I check all modifications: there are many manual conversion from a byte offset to a block ID.

```rust
let block_id = BlockId::new(offset / BLOCK_SIZE);
```

But no usage of `from_offset` is found at all. This result seems to suggest that it is quite common to convert an _unaligned_ offset to a block ID and this is considered harmless.

So I would suggest having two methods to convert an offset to a block ID, one is infallible, the other is fallible.

```rust
pub fn from_byte_offset(offset: usize) -> Self {
    Self(offset / BLOCK_SIZE)
}

pub fn from_byte_offset_aligned(offset: usize) -> Result<Self> {
    Self::check_align(offset)?;
    Ok(Self((offset / BLOCK_SIZE) as _))
}
```

And `from_byte_offset` should be the one that is most often used.

tatetian (commented):
Thanks for submitting the updated version. Here are my comments for the first commit. I will submit the comments for the second commit later.

src/libos/crates/block-device/src/block_id.rs
lucassong-mh: My first thought about these two private type alias: 1. more readable 2. if we want to change the type, we just modify one place. And what you comment below convinces me.

src/libos/crates/block-device/src/block_id.rs
lucassong-mh: I think it's strange in the first place, but would `Add<u64>` be ambiguous to user? (whether i should add a `Offset` or `RawBlockId`?)

src/libos/crates/block-device/src/block_id.rs
lucassong-mh: So do we still need a `new()`(or rename to `from_raw()`) to take `raw_bid: RawBId` as input, or just use two infallible/fallible fns that only takes `offset: usize`

src/libos/crates/block-device/src/block_id.rs
tatetian: `u64` seems to be the best type to represent the block numbers. `Offset` is for byte offsets, while `RawBlockId` (although) is yet another kind of IDs. Neither byte offsets or another type of IDs should be added to block IDs. So I think implementing `Add<u64>` for `BlockId` is our best choice.

src/libos/crates/block-device/src/block_id.rs
tatetian: We need to keep `new`, which should be viewed as the default constructor for `BlockId`.

src/libos/crates/jindisk/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+Both implementations are being developed actively. They are ready for technical preview, but still lack some production-grade features.
+
+Our long-term plan is to ultimately merge the two implementations into one unified Rust implementation that can be integrated with Linux, Occlum, and probably other OSes as well. We will submit patches to the Linux community so that JinDisk may be eventually included in the mainline Linux.
+
+For more information about the design of JinDisk, see [the paper]().
+
+## How to Use
+
+Current crate is the Occlum version of JinDisk. To use the crate, follow the usage example in crate's public docs. To test basic functions, run `cargo test` under current crate. To further evaluate JinDisk's performance, follow the steps in benchmark demos of Occlum, [FIO](../../../../demos/benchmarks/fio/) or [Filebench](../../../../demos/benchmarks/filebench/), test JinDisk under directory `/sfs` or `/dev/jindisk`.
```
tatetian: Should start with "The current crate", not "Current crate".

src/libos/crates/block-device/src/util/block_range.rs
```diff
@@ -30,9 +30,9 @@ impl Iterator for BlockRangeIter {
 
         // Construct sub-range of next block
         let sub_range = {
-            let block_id = self.begin / self.block_size;
+            let block_id = Bid::from_raw((self.begin / self.block_size) as _);
```
tatetian: Use `Bid::from_byte_offset`.

src/libos/crates/block-device/src/block_id.rs
```diff
@@ -0,0 +1,66 @@
...
+            "block id is too big"
+        );
+        Self(raw_bid)
+    }
+
+    pub fn to_raw(self) -> RawBid {
+        self.0
+    }
+
+    pub fn from_byte_offset(offset: usize) -> Self {
```
tatetian: This is also a const function.

src/libos/crates/block-device/src/block_id.rs
```diff
@@ -0,0 +1,66 @@
...
+use crate::prelude::*;
+use errno::return_errno;
+
+pub type RawBid = u64;
+
+#[derive(Copy, Clone, Debug, Hash, PartialEq, Eq, PartialOrd, Ord)]
+pub struct Bid(RawBid);
+
+impl Bid {
+    pub const fn from_raw(raw_bid: RawBid) -> Self {
```
tatetian: Using `from_raw` as the name is ok. But I think `new` is a little bit better because this method should be viewed as the constructor (not a conversion method), which is usually named `new`. It is up to you which one to choose.

src/libos/crates/block-device/src/block_id.rs
```diff
@@ -0,0 +1,66 @@
...
+
+    fn check_align(offset: usize) -> Result<()> {
+        if offset % BLOCK_SIZE != 0 {
+            return_errno!(EINVAL, "offset not aligned with block size")
+        }
+        Ok(())
+    }
+}
+
+impl Add<RawBid> for Bid {
```
tatetian: As have I explained in one of the responses to the original comment, `Add<u64>` is semantically better than `Add<RawBid>`, although the two are essentially the same.

src/libos/crates/jindisk/src/lib.rs
```diff
@@ -0,0 +1,107 @@
...
+//! // Read block contents from JinDisk
+//! let mut rbuf = [0u8; BLOCK_SIZE];
+//! jindisk.read(offset, &mut rbuf).await?;
+//!
+//! // Sync all data to storage medium
+//! jindisk.sync().await?;
+//!
+//! assert_eq!(rbuf, wbuf);
+//! ```
+//!
```
tatetian, line 61: This is a good crate-level documentation with instructive examples 👍

src/libos/crates/jindisk/src/jindisk.rs
```diff
@@ -0,0 +1,508 @@
...
+    superblock: SuperBlock,
+    data_cache: Arc<DataCache>,
+    lsm_tree: Arc<LsmTree>,
+    checkpoint: Arc<Checkpoint>,
+    cleaner: Cleaner,
+    root_key: Key,
+}
+
+impl JinDisk {
+    /// Create a new `JinDisk`, initialize basic structures.
```
tatetian, line 59: Good effort in writing documents for all public APIs 👍

src/libos/crates/jindisk/src/jindisk.rs
```diff
@@ -0,0 +1,508 @@
...
+        );
+        self.superblock
+            .persist(
+                &disk_superblock_view,
+                SUPER_BLOCK_REGION_ADDR,
+                &self.root_key,
+            )
+            .await?;
+
+        self.lsm_tree.persist().await?;
```
tatetian: What if this line or the lines below returns error? In such cases, the update to the superblock is already persisted, but the ones to other on-disk data structures are not. We need to deal with such situations---eventually, in the future, not this commit. Consider adding a TODO.

This is just one example of the classic mistake in error handling: partial results (e.g., data persistence) are not rolled back after an error occurs. In Rust, this kind of mistakes are particularly easy to make as returning to the caller directly upon errors is a common practice. We will need to review the rest of the source code and see if there are more such mistakes.

src/libos/crates/jindisk/src/index/mem_table.rs
```diff
@@ -0,0 +1,214 @@
...
+/// A thread-safe `MemTable`.
+pub struct LockedMemTable {
+    mem_table: AsyncRwLock<BTreeMemTable>,
+    num_records: AtomicUsize,
+    capacity: usize,
+    reclaim_fn: Option<ReclaimFn>,
+}
+
+pub type ReclaimFn =
+    Box<dyn Fn(Record, MemRecord) -> Pin<Box<dyn Future<Output = MemRecord> + Send>> + Send + Sync>;
```
tatetian: Is it possible not to copy records? In other words, can we change the signature to the following form.

```rust
pub type ReclaimFn =
    Box<dyn Fn(&Record, &MemRecord) -> Pin<Box<dyn Future<Output = MemRecord> + Send>> + Send + Sync>;
```

tatetian (commented):
Well done! I really appreciate your persevere and patience in improving the code quality constantly. The code is now good enough for merge. Before we do, let's fix some trivial issues listed in the comments below. Then, you can re-submit this PR to the Occlum repo (not NGO). See you there!

src/libos/crates/jindisk/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+Prior disk I/O protection solutions only provide a subset of JinDisk's security guarantees. For example, Linux's [dm-crypt](https://docs.kernel.org/admin-guide/device-mapper/dm-crypt.html) and [dm-integrity](https://docs.kernel.org/admin-guide/device-mapper/dm-crypt.html) only protect confidentiality and integrity, respectively. Although Linux's [dm-verity](https://docs.kernel.org/admin-guide/device-mapper/verity.html) ensures both integrity and freshness, it is read-only. As another example, [Intel SGX Protected File System Library](https://www.intel.com/content/www/us/en/developer/articles/technical/overview-of-intel-protected-file-system-library-using-software-guard-extensions.html) protects confidentiality, integrity, freshness, and consistency, but falls short of atomicity and anonymity.
+
+## Implementations
+
+As of this moment, JinDisk has two implementations.
+
+* The Linux version ([repo](https://github.com/jinzhao-dev/jindisk)), written in C, is for use in VM TEEs like [AMD SEV](https://developer.amd.com/sev/) and [Intel TDX](https://www.intel.com/content/www/us/en/developer/articles/technical/intel-trust-domain-extensions.html).
```
tatetian: The URL for the Linux version has changed.

PR #346, opened 2022-12-07
[libos] Create and assign an io_uring instance for each vcpu

jessehui:


src/libos/src/entry/enclave.rs
```diff
@@ -221,7 +213,14 @@ pub extern "C" fn occlum_ecall_run_vcpu(pal_data_ptr: *const occlum_pal_vcpu_dat
     assert!(check_ptr(pal_data_ptr).is_ok()); // Make sure the ptr is outside the enclave
     set_pal_data_addr(pal_data_ptr);
 
-    let running_vcpu_num = async_rt::executor::run_tasks();
+    // Init an io_uring instance for this vcpu
```
volcano0dr, line 216: I think it is possible to use a static `COUNTER` inside the function as the index of `IO_URING_MANAGER`, so that there is no need to modify the interface of async-rt.

`static COUNTER: AtomicU32`

PR #351, opened 2022-12-12
Pick some commits

liqinggd:
1. [Print the raw errno if it is not an expected value](https://github.com/occlum/ngo/commit/8e2765a92f66b7a0b6a8c678e1c20dbadf1c0c3f)
2. [Improve the usability of occlum start/exec/stop cmd](https://github.com/occlum/ngo/commit/32cc2d027277afcd95acacf1ca5db141eced979c)


volcano0dr (approved):
LGTM

